PR #2, opened 2018-07-24
Clean up helper scripts

johnrevans6:
Performed a general cleanup of the setup and cleanup helper scripts. 
Adding checks to make sure a user is passing a valid project id and service account name into the helper scripts.
Fixed a bug where the credentials.json file wasn't being deleted.


helpers/cleanup.sh
```diff
@@ -14,12 +14,25 @@
 # limitations under the License.
 
 
-PROJECT_ID="$1"
+PROJECT_ID="$(gcloud projects list --format="value(projectId)" | grep -Eo "$1")"
```
morgante: If I'm reading this correctly, we default to using the gcloud project ID? Instead I think any passed project ID should take precedence.

helpers/cleanup.sh
```diff
@@ -14,12 +14,25 @@
...
+then
+    echo "ERROR The specified project wasn't found."
+    exit 1;
+fi
+
 ORG_ID="$(gcloud projects describe ${PROJECT_ID} --flatten=parent.id | grep -Eo "\d+")"
-SERVICE_ACCOUNT_ID="$2"
-STAGING_DIR="$(echo ${PWD}/staging)"
+SERVICE_ACCOUNT_ID="$(gcloud iam service-accounts list --format="value(email)" | grep -Eo "$2@${PROJECT_ID}.iam.gserviceaccount.com")"
+KEY_FILE="$(echo ${PWD}/credentials.json)"
```
morgante: I don't think echo is needed here. `KEY_FILE="${PWD}/credentials.json"` should do the same?

helpers/setup.sh
```diff
@@ -14,15 +14,20 @@
...
-ORG_ID="$(gcloud projects describe $PROJECT_ID --format="value(parent.id)")"
+PROJECT_ID="$(gcloud projects list --format="value(projectId)" | grep -Eo "$1")"
+
+if [[ $PROJECT_ID == "" ]];
+then
+    echo "ERROR The specified project wasn't found."
+    exit 1;
+fi
+
+ORG_ID="$(gcloud projects describe ${PROJECT_ID} --format="value(parent.id)")"
```
morgante: What about nested projects? Parent might be a folder.

helpers/cleanup.sh
johnrevans6: Ah, you are correct.

helpers/cleanup.sh
johnrevans6: This line basically checks to see if the project exists. First, we get a list of projects for the authenticated user, and then there's a check to see if the project the user has supplied exists. If it doesn't, then an error gets thrown. So the project id that is being used is whatever the user is providing.

helpers/setup.sh
johnrevans6: Good point. I'll need to think on this.

johnrevans6: Hmmm maybe I overthought this one. My initial thinking was that structurally, folders are considered to be suborganizations. So you can have one parent org, and then multiple suborgs. I wonder if it makes sense to just do something like...

gcloud organizations list --format="value(ID)" 

to grab the parent organization? Otherwise, might need to rely on the user here, and then supply an additional check to validate the org exists.

helpers/cleanup.sh
johnrevans6: Actually, this is cleaner:

PROJECT_ID="$(gcloud projects list --format="value(projectId)" --filter=$1)"

helpers/setup.sh
morgante: Opened this to address it later: #3

PR #4, opened 2018-07-31
Activate Services in Terraform Instead of in Forseti

johnrevans6:
Forseti has several service dependencies that should be managed by Terraform. Pulled these dependencies in to the module. Note, calling destroy on the Compute Engine API when there are still resources in use will cause terraform to fail to deactivate the service.


examples/simple/variables.tf
```diff
@@ -34,3 +34,14 @@ variable "project_id" {
 variable "credentials_file_path" {
   description = "Path to service account json"
 }
+
```
morgante: @johnrevans6 Looks like some unrelated changes got committed.

main.tf
```diff
@@ -31,8 +44,18 @@ locals {
...
+  Activate services
  *******************************************/
+resource "google_project_service" "activate_services" {
+  count   = "${length(local.services_list)}"
+  project = "${local.project_id}"
+
+  service = "${element(local.services_list, count.index)}"
+}
+
+# /*******************************************
```
morgante: Why is the `#` added on the beginning of these lines? Please remove.

outputs.tf
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
```
morgante: Please keep this blank line.

examples/simple/outputs.tf
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+
```
morgante: Looks like a whitespace error got added. This should be an empty blank line.

outputs.tf
johnrevans6: ack

PR #12, opened 2018-09-24
Correcting service-account script, adding a script to automate the in…

lbastet:
To fully automate the installation with Terraform I create a script that changes to lines of Forseti Security installer that wants to received a raw_input from the keyboard to get the var of variable.tf regarding the email owner account. On the other hand, I correct a mistake of the helper/setup.sh script of double quotes that was wrong place. Best regards

### DIFF
--- a/accScript.sh
+++ b/accScript.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# Author: luciano.bastet@globant.com                        #
+# This script replace raw_input for the owner email account #
+# in the Forsety Security installer Script to automate      #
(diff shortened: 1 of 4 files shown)


johnrevans6:
Looks good! I'm still not clear what changed on the Forseti side. So I'll be investigating this when I get some time, but this will indeed fix the issues you observed.

lbastet:
Hi johnrevans6 I have a question. If i have to many projects for the Organization do I have to run de helper/setup.sh for each project? I need a service account at org and project level for each project? Do I need to install Forseti for each project?

PR #22, opened 2018-11-15
refer to accScript.sh from module path, not root path.

bmenasha:
Otherwise when terraform-google-forseti is used as a module accScript.sh
execution fails. (also it's no longer in the scripts/ directory.


morgante:
This should be moved into the `scripts/` directory. Do you mind doing that as part of the PR?

PR #37, opened 2019-01-23
Remove obsolete forseti-setup.tf

adrienthebo:
This terraform file is old and somewhat dangerous, as it manages the
full IAM binding on organization and project resources. The
`helpers/setup.sh` script does this more safely, so we no longer need
this file.


ocervell:
LGTM

PR #48, opened 2019-02-11
Implement host integration tests

adrienthebo:
The current test coverage validated that Terraform had created the required Forseti resources but did not validate the Forseti installation itself. If Forseti failed to install for any reason the integration tests would still pass, which indicated a big gap in our test coverage.

This pull request adds test coverage on the Forseti client and server instances.  This coverage allows us to confirm that the Forseti server is operational and the Forseti client can connect to the server.


test/integration/simple_example/controls/forseti.rb
```diff
@@ -121,26 +108,17 @@
...
 
-control 'forseti-server-service-account' do
-  title 'Test Forseti Server Service Account'
   describe google_service_account(name: "projects/#{project_id}/serviceAccounts/#{forseti_server_service_account}") do
     its(:email) { should eq forseti_server_service_account }
     its(:display_name) { should eq "Forseti Server Service Account" }
   end
-end
 
-control 'forseti-firewall-rules' do
```
morgante: Why are we combining all controls into one?

adrienthebo: In order to target GCP and the hosts we need to assign specific controls to specific systems (https://github.com/terraform-google-modules/terraform-google-forseti/pull/48/files#diff-d3fbd21bc6be7ce4531c27a82b2619deR29); without this assignment we'll attempt to run the client and server controls with the GCP backend, which will fail. We could break out the controls as was done previously and add the list to `.kitchen.yml` - but we're not getting value from having fine grained controls. Pushing all of the example groups into a single control provides the functionality we need and doesn't have any real cost, aside from slightly different output formatting.

morgante: Got it, thanks.

PR #50, opened 2019-02-11
Parameterize and add KMS rule file

hshin-g:
- Parameterize all user configurable values in forseti_conf_server.yaml.tpl
- Add in KMS scanner rule file into rules directory


modules/server/templates/configs/forseti_conf_server.yaml.tpl
```diff
@@ -374,8 +376,28 @@ notifier:
                 # gcs_path should begin with "gs://"
                 gcs_path: gs://${FORSETI_BUCKET}/scanner_violations
 
-        - resource: groups_violations
+        - resource: kms_violations
           should_notify: true
+          notifiers: ${KMS_VIOLATIONS_SHOULD_NOTIFY}
+            # Email violations
```
adrienthebo: It looks like there's a two space indentation here which will cause this to be invalid YAML

hshin-g: Resolved- thanks Adrien.

variables.tf
```diff
@@ -79,6 +79,458 @@ variable "server_instance_metadata" {
...
+
+#-------------------------------#
+# Forseti server config scanner #
+#-------------------------------#
+variable "audit_logging_enabled" {
+  description = "Audit Logging scanner enabled."
+  default     = "false"
+}
+
+variable "biquery_enabled" {
```
adrienthebo: Typo - this should be `bigquery_enabled`, we dropped the first g in `bigquery`

hshin-g: Good eye. Changed.

PR #56, opened 2019-02-14
Implement CI

adrienthebo:
This pull request implements scripts and functionality for verifying this module within the CFT Concourse CI system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ DOCKER_REPO_BASE_KITCHEN_TERRAFORM 	:= ${DOCKER_ORG}/cft/kitchen-terraform:${DOC
 # All is the first target in the file so it will get picked up when you just run 'make' on its own
 all: check_shell check_python check_golang check_terraform check_docker check_base_files test_check_headers check_headers check_trailing_whitespace generate_docs
 
+.PHONY: check
+check: check_shell check_python check_golang check_terraform check_docker check_base_files test_check_headers check_headers check_trailing_whitespace
+
 # The .PHONY directive tells make that this isn't a real target and so
 # the presence of a file named 'check_shell' won't cause this target to stop
 # working
--- a/Makefile
+++ b/Makefile
@@ -75,12 +75,7 @@ check_headers:
 # Integration tests
 .PHONY: test_integration
 test_integration:
-	bundle install
-	bundle exec kitchen create
-	bundle exec kitchen converge
-	bundle exec kitchen converge
-	bundle exec kitchen verify
-	bundle exec kitchen destroy
+	test/ci_integration.sh
 
 .PHONY: generate_docs
 generate_docs:
--- a/Makefile
+++ b/Makefile
@@ -69,8 +69,7 @@ test_check_headers:
 
 .PHONY: check_headers
 check_headers:
-	@echo "Checking file headers"
-	@python test/verify_boilerplate.py
+	@source test/make.sh && check_headers
 
 # Integration tests
 .PHONY: test_integration
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,11 @@ version:
 .PHONY: docker_run
 docker_run:
 	docker run --rm -it \
+		-e PROJECT_ID \
+		-e ORG_ID \
+		-e DOMAIN \
+		-e GSUITE_ADMIN_EMAIL \
+		-e SERVICE_ACCOUNT_JSON \
 		-e CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=${CREDENTIALS_PATH} \
 		-e GOOGLE_APPLICATION_CREDENTIALS=${CREDENTIALS_PATH} \
 		-v $(CURDIR):/cft/workdir \
@@ -98,6 +103,11 @@ docker_run:
 .PHONY: docker_create
 docker_create:
 	docker run --rm -it \
+		-e PROJECT_ID \
+		-e ORG_ID \
+		-e DOMAIN \
+		-e GSUITE_ADMIN_EMAIL \
+		-e SERVICE_ACCOUNT_JSON \
 		-e CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=${CREDENTIALS_PATH} \
 		-e GOOGLE_APPLICATION_CREDENTIALS=${CREDENTIALS_PATH} \
 		-v $(CURDIR):/cft/workdir \
@@ -107,6 +117,11 @@ docker_create:
 .PHONY: docker_converge
 docker_converge:
 	docker run --rm -it \
+		-e PROJECT_ID \
+		-e ORG_ID \
+		-e DOMAIN \
+		-e GSUITE_ADMIN_EMAIL \
+		-e SERVICE_ACCOUNT_JSON \
 		-e CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=${CREDENTIALS_PATH} \
 		-e GOOGLE_APPLICATION_CREDENTIALS=${CREDENTIALS_PATH} \
 		-v $(CURDIR):/cft/workdir \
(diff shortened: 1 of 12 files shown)


test/make.sh
```diff
@@ -14,83 +14,145 @@
...
+    fi
   done
-  rm -f $TMPFILE
+}
+
+function check_headers() {
+  echo "Checking file headers"
+  # Use the exclusion behavior of find_files
+  find_files . -type f -print0 \
+    | grep -v 'modules/rules/*.yaml' \
```
adrienthebo: Calling this out - I'm specifically ignoring the Forseti rules files since forseti-security project holds the copyright. Is this a good thing to do?

test/fixtures/simple_example/main.tf
```diff
@@ -37,3 +37,35 @@ resource "local_file" "gce-keypair-pk" {
...
   }
 }
+
+resource "null_resource" "wait_for_server" {
+  triggers = {
+    server_instance_id = "${uuid()}"
+  }
+
+  provisioner "remote-exec" {
+    inline = "while ! [ -f /etc/profile.d/forseti_environment.sh ]; do sleep 10; done"
```
adrienthebo: Calling this out - this approach does work but is a bit crude. Is it worth extracting this to a script?

test/ci_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+set +x
+echo "${SERVICE_ACCOUNT_JSON}" > "${CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE}"
+set -x
+GOOGLE_APPLICATION_CREDENTIALS="${CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE}"
+
+export TF_VAR_project_id="${PROJECT_ID}"
+export TF_VAR_org_id="${ORG_ID}"
+export TF_VAR_domain="${DOMAIN}"
+export TF_VAR_gsuite_admin_email="${GSUITE_ADMIN_EMAIL}"
+export TF_VAR_credentials_path="${CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE}"
```
jeffmccune: -1 Wrap all of the environment setup behavior into a function.  The reason is to speed up smoke testing and code reviews with `source test/ci_integration.sh && setup_environment` from within `make docker_run`.

See example: https://github.com/terraform-google-modules/terraform-google-startup-scripts/blob/3a904a59468c427a93d0e656005836f4a88464a2/test/ci_integration.sh#L35

test/ci_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+
+declare -rx CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE GOOGLE_APPLICATION_CREDENTIALS
+
+set -e
+bundle install
```
jeffmccune: Don't need to block this PR, but the docker image already contains kitchen-terraform 4.1.0.

Suggest adjusting the Gemfile to ~> 4.1.0 and removing the Gemfile.lock to speed up the tests and avoid installing ~100+ dependencies from rubygems.org every run

test/fixtures/simple_example/main.tf
jeffmccune: I generally put this kind of stuff into a file resource if only to get automated shellcheck and such.

Regarding the logic, sleep 10 seems like a long time.  The stat call is cheap, could tighten this up to 1 second or shorter.

Finally, it's an infinite loop.  Suggest adding some upper bound and a log message to give the user some feedback about what's going on.

test/make.sh
jeffmccune: -2 Regarding correctness, I think this skips all headers.  That grep looks like a file glob instead of a regular expression and print0 is sending null-separated output.  Does grep handle that properly?

Regarding the headers themselves, I think it's OK to skip them.  Would be better if we could check them somehow, but maybe as a follow up PR.

Makefile
```diff
@@ -66,18 +69,12 @@ test_check_headers:
...
 # Integration tests
 .PHONY: test_integration
 test_integration:
-	bundle install
-	bundle exec kitchen create
-	bundle exec kitchen converge
-	bundle exec kitchen converge
-	bundle exec kitchen verify
-	bundle exec kitchen destroy
+	test/ci_integration.sh
```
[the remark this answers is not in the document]
adrienthebo: Resolved in a4119267a44a585b0978f1308907474becc87464

test/ci_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -x
+
+if [ -z "${PROJECT_ID}" ]; then
```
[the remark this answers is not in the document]
adrienthebo: Resolved in 4cbf775fe515a26a4719239a92e19d48664ad4be

test/ci_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+	echo "The PROJECT_ID ENV variable must be set to proceed. Aborting."
+	exit 1
+fi
+
+if [ -z "${ORG_ID}" ]; then
+	echo "The ORG_ID ENV variable must be set to proceed. Aborting."
+	exit 1
```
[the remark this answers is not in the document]
adrienthebo: Resolved in 4cbf775fe515a26a4719239a92e19d48664ad4be

test/ci_integration.sh
adrienthebo: Resolved in 4cbf775fe515a26a4719239a92e19d48664ad4be

test/fixtures/simple_example/main.tf
adrienthebo: Good point about shellcheck, resolved in c6faeb9606d3934b6f1addabbe97a1f9f75222d9

test/ci_integration.sh
adrienthebo: Sounds good - though what's the reasoning for pinning to `~> 4.1.0` over `~> 4.1`?

test/make.sh
adrienthebo: I've revamped the logic for filtering out Forseti rules files, could you check that out?

test/ci_integration.sh
aaron-lane: I believe this was already raised as an issue to address within the Docker image.

test/fixtures/simple_example/main.tf
```diff
@@ -37,3 +37,35 @@ resource "local_file" "gce-keypair-pk" {
     sshKeys = "ubuntu:${tls_private_key.main.public_key_openssh}"
   }
 }
+
+resource "null_resource" "wait_for_server" {
+  triggers = {
+    server_instance_id = "${uuid()}"
```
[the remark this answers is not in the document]
adrienthebo: Resolved in https://github.com/terraform-google-modules/terraform-google-forseti/pull/56/commits/847ed6391d5ed4a2203f3746e1023a623c71cf07

test/fixtures/simple_example/main.tf
```diff
@@ -37,3 +37,35 @@ resource "local_file" "gce-keypair-pk" {
...
+      user        = "ubuntu"
+      host        = "${module.forseti-install-simple.forseti-server-vm-public-ip}"
+      private_key = "${tls_private_key.main.private_key_pem}"
+    }
+  }
+}
+
+resource "null_resource" "wait_for_client" {
+  triggers = {
+    client_instance_id = "${uuid()}"
```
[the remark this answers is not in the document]
adrienthebo: Resolved in https://github.com/terraform-google-modules/terraform-google-forseti/pull/56/commits/847ed6391d5ed4a2203f3746e1023a623c71cf07

PR #61, opened 2019-02-26
Config/rule file changes for Forseti v2.12.0

hshin-g:
Config/rule file changes for Forseti v2.12.0

- Edit KMS rule file (PR: https://github.com/GoogleCloudPlatform/fors…eti-security/pull/2584)
- Add config section for composite root (PR: https://github.com/GoogleCloudPlatform/forseti-security/pull/2538)

### DIFF
--- a/modules/server/templates/configs/forseti_conf_server.yaml.tpl
+++ b/modules/server/templates/configs/forseti_conf_server.yaml.tpl
@@ -6,9 +6,31 @@ global:
 
 inventory:
 
+    # You must set ONLY one of root_resource_id or composite_root_resources in
+    # your configuration. Defining both will cause Forseti to exit with an
+    # error.
+
     # Root resource to start crawling from, formatted as
     # <resource_type>/<resource_id>, (e.g. "organizations/12345677890")
     root_resource_id: ${ROOT_RESOURCE_ID}
+
+    # Composite root resources: combines multiple resource roots into a single
+    # inventory, for use across all Forseti modules. Can contain one or more
+    # resources from the GCP Resource Hierarchy in any combination.
+    # https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy
+    #
+    # All resources must grant the appropriate IAM permissions to the Forseti
+    # service account before they can be included in the inventory.
+    #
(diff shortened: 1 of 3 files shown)


modules/server/templates/configs/forseti_conf_server.yaml.tpl
```diff
@@ -6,10 +6,32 @@ global:
...
+    # https://cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy
+    #
+    # All resources must grant the appropriate IAM permissions to the Forseti
+    # service account before they can be included in the inventory.
+    #
+    # Forseti Explain is not supported with a composite root at this time.
+    #
+    # Resources can exist in multiple organizations.
+    #
+    #composite_root_resources:
```
adrienthebo: This setting will not be user-tunable; if a user hand-edits this file with a composite root resource then Terraform will replace it the next time Terraform runs. Is this desirable?

variables.tf
```diff
@@ -27,7 +27,7 @@ variable "gsuite_admin_email" {
 
 variable "forseti_version" {
   description = "The version of Forseti to install"
-  default     = "v2.11.1"
+  default     = "v2.12.0"
```
adrienthebo: Since forseti-security hasn't tagged `v2.12.0` this has uncovered a bug - if the `forseti_version` can't be checked out it looks like we'll just check out the dev branch. Should the Forseti installation abort if an invalid version is provided?

hshin-g: Good point- installation should abort if an invalid version is provided.

modules/server/templates/configs/forseti_conf_server.yaml.tpl
hshin-g: Would prefer if this was user-tunable, but was not sure of the best way to incorporate this, as it could potentially be multiple lines and can vary from user to user. Thoughts on how to address?

adrienthebo: We can do this, but it'll be fairly involved - I think that we can address this in a followup PR so that forseti-security v2.12.0 can land in time.

PR #62, opened 2019-02-26
Add @adrienthebo and @aaron-lane to CODEOWNERS

adrienthebo:
When we originally added the CODEOWNERS file we only included Morgante, since CI was not yet setup. Now that we have functioning CI we can broaden the number of code reviewers for this module.


morgante:
Approved and set up branch protection on `master`.

PR #64, opened 2019-02-27
- Add KMS rule file

hshin-g:
Add KMS scanner rule file


adrienthebo:
I've added test coverage for this in https://github.com/terraform-google-modules/terraform-google-forseti/pull/65, we'll need to merge this first and then CI should turn green for that PR.

adrienthebo:
CI is testing the wrong pull request (#65 instead of #64), which means that the checks on this PR are failing incorrectly. I'm verifying locally and will merge if it passes.

adrienthebo:
Verified that the `simple-example` kitchen configuration correctly populates `rules/kms_rules.yaml`.

PR #66, opened 2019-02-28
Update setup.sh and cleanup.sh

marko7460:
This PR fixes setup.sh and cleanup.sh to:
1. Handle shared-vpc use case
2. Update roles on SA account
3. Enable necessary services on project in order for terraform to run without problems

### DIFF
--- a/helpers/cleanup.sh
+++ b/helpers/cleanup.sh
@@ -13,8 +13,63 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Usage info
+show_help() {
+cat << EOF
+Usage: ${0##*/} [-p PROJECT_ID] [-f HOST_PROJECT] [-o ORG_ID]
+Remove service account created by setup.sh script.
 
-PROJECT_ID="$(gcloud projects list --format="value(projectId)" --filter="$1")"
+    -p PROJECT_ID         project id where service account will be created.
+    -s SERVICE ACCOUNT    service account name.
+    -f HOST_PROJECT       id of a project holding shared vpc.
+    -o ORG_ID             organization id.
+    -h                    this help.
+Example: ./cleanup.sh -p test-project-1 -s cloud-foundation-forseti -o organization.com
+EOF
+}
+
+# Initialize opt variables:
+project_id=""
+host_project=""
+service_account_id=""
+org=""
+
+OPTIND=1
+while getopts ":h:p:s:f:o:" opt; do
+    case $opt in
+        h)
+            show_help
+            exit 0
+            ;;
+        p)  project_id=$OPTARG
+            ;;
+        s)  service_account_id=$OPTARG
+            ;;
+        f)  host_project=$OPTARG
+            ;;
+        o)  org=$OPTARG
+            ;;
+        *)
+            show_help >&2
+            exit 1
+            ;;
+    esac
+done
+shift "$((OPTIND-1))"   # Discard the options and sentinel --
+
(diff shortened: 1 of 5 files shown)


adrienthebo:
We're about to release v1.2.0 of this module to match to the forseti-security v2.12.0 release; in the interest in minimizing risk to this release I'm going to defer this until v1.2.1 or v1.3.0. My quick review of this is positive, will take it for a more thorough spin after v1.2.0 is out. Thanks!

helpers/cleanup.sh
```diff
@@ -13,8 +13,60 @@
...
+            show_help >&2
+            exit 1
+            ;;
+    esac
+done
+shift "$((OPTIND-1))"   # Discard the options and sentinel --
+
+if [[ $project_id == "" ]];
+then
+    echo "ERROR -p option is mandatory"
```
aaron-lane: An option being mandatory is a contradiction. Mandatory arguments should be positional.

helpers/cleanup.sh
```diff
@@ -13,8 +13,60 @@
...
+    exit 1
+fi
+
+if [[ $service_account_id == "" ]];
+then
+    echo "ERROR -s option is mandatory"
```
aaron-lane: Ditto about mandatory options.

helpers/setup.sh
```diff
@@ -13,18 +13,63 @@
...
+            show_help >&2
+            exit 1
+            ;;
+    esac
+done
+shift "$((OPTIND-1))"   # Discard the options and sentinel --
+
+if [[ $project_id == "" ]];
+then
+    echo "ERROR -p option is mandatory"
```
aaron-lane: Ditto about mandatory options.

helpers/setup.sh
```diff
@@ -35,6 +80,21 @@ SERVICE_ACCOUNT_ID="${SERVICE_ACCOUNT_NAME}@${PROJECT_ID}.iam.gserviceaccount.co
 STAGING_DIR="${PWD}"
 KEY_FILE="${STAGING_DIR}/credentials.json"
 
+if [[ ${host_project} != "" ]];
```
aaron-lane: Logic which is duplicated between `setup.sh` and `cleanup.sh` should be extracted to a shared library file.

adrienthebo: I'm normally a big proponent for DRY, but for things like simple shell scripts I tend to relax that a bit. Using a shared library reduces the amount of duplicated code but incurs the complexity cost of loading that shared library, which can be pretty obnoxious depending on the bash version and whatnot. We only have two copies of this logic, so I think that we should push back on this refactoring. If we need in the future let's add it, but let's wait for the demand to present itself.

helpers/cleanup.sh
adrienthebo: Argument parsing styles are pretty subjective; I've seen (and written) a number of CLI tools that have used both positional arguments and required flags. Requiring flags for values is slightly more verbose but increases readability, which seems like a decent tradeoff.

In addition argument parsing in bash is generally less than pleasant, and mixing positional arguments with flags is generally a nightmare. For the sake of simplicity and readability, I actually prefer the way that we have this now. Is there a cost to this approach that I'm not seeing?

helpers/setup.sh
adrienthebo: After additional review I've seen small issues that affect option parsing in both setup and cleanup; this makes me shift a bit closer towards having a shared file containing common definitions.

PR #67, opened 2019-02-28
Feature/change source_ranges

marko7460:
This PR transforms `source_ranges` variable into three variables necessary for firewall rules on the server and the client. Variables are:
1. `server_grpc_allow_ranges`
2. `server_ssh_allow_ranges`
3. `client_ssh_allow_ranges`

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -79,10 +79,16 @@ variable "server_instance_metadata" {
   default     = {}
 }
 
-variable "source_ranges" {
+variable "server_grpc_allow_ranges" {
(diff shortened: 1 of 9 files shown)


variables.tf
```diff
@@ -561,6 +567,12 @@ variable "client_instance_metadata" {
   default     = {}
 }
 
+variable "client_ssh_allow_ranges" {
+  description = "List of CIDRs that will be allowed ssh access to forseti server"
```
adrienthebo:
```suggestion
  description = "List of CIDRs that will be allowed ssh access to forseti client"
```

PR #71, opened 2019-03-01
Adding bigquery.metadataViewer role to SA

marko7460:
Adding `roles/bigquery.metadataViewer` to SA as requested by @hshin-g

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -81,14 +81,17 @@ suites:
             - forseti-server
             - forseti-client
             - forseti-shared-firewall
-        - name: shared_vpc
+            - forseti
+        - name: gcloud
           backend: local
           controls:
             - forseti-subnetwork
+            - forseti-org-iam
         - name: server
           backend: ssh
           controls:
             - forseti-command-server
+            - server
           hosts_output: forseti-server-vm-public-ip
           user: ubuntu
           key_files:
@@ -99,6 +102,7 @@ suites:
           backend: ssh
           controls:
             - forseti-command-client
+            - client
           hosts_output: forseti-client-vm-public-ip
           user: ubuntu
           key_files:
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -54,6 +54,12 @@ suites:
             - test/fixtures/simple_example/sshkey
           shell: true
           shell_options: "--login"
+
+        - name: org-iam
+          backend: local
+          controls:
+            - forseti-org-iam
+
       provisioner:
         name: terraform
 
--- a/test/integration/simple_example/controls/forseti.rb
+++ b/test/integration/simple_example/controls/forseti.rb
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+org_id = attribute('org_id')
 project_id = attribute('project_id')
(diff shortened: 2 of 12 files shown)


test/integration/simple_example/controls/forseti.rb
```diff
@@ -184,5 +185,38 @@
       end
     end
   end
+end
+
+control 'forseti-org-iam' do
```
adrienthebo: 💯 , thanks for adding this test coverage!

.kitchen.yml
```diff
@@ -54,6 +54,12 @@ suites:
             - test/fixtures/simple_example/sshkey
           shell: true
           shell_options: "--login"
+
+        - name: org-iam
```
adrienthebo: Right now these tests will only run on the `simple-example` test fixture, which means we have a gap for shared-vpc. We should address this but that change will be somewhat involved, so this is mainly a note for posterity.

test/integration/simple_example/controls/forseti.rb
```diff
@@ -184,5 +185,38 @@
...
+  title "Validate organization roles of SA"
+  describe command("gcloud organizations get-iam-policy #{org_id} --filter='bindings.members:#{forseti_server_service_account}' --flatten='bindings[].members' --format='csv(bindings.role)'") do
```
adrienthebo: In most/all of our tests that shell out to `gcloud` we generally use the `json` format. Is there a particular reason why we chose CSV for this test?

marko7460: It was easier for me to list the roles in CSV and parse in ruby. https://github.com/marko7460/terraform-google-forseti/blob/feature/add_missing_role/test/integration/simple_example/controls/forseti.rb#L196-L201.
I can fix the code to deal with json if that is important for consistency with other projects and tests.

test/integration/simple_example/controls/forseti.rb
```diff
@@ -184,5 +185,38 @@
...
+    its(:exit_status) { should eq 0 }
+    its(:stderr) { should eq '' }
+
+    let(:sa_roles) do
```
aaron-lane: This definition can be simplified.

```rb
let :sa_roles do
  CSV.parse(subject.stdout)[1..-1].map(&:first)
end
```

test/integration/simple_example/controls/forseti.rb
```diff
@@ -184,5 +185,38 @@
...
+      roles = Array.new
+      CSV.parse(subject.stdout).each {|role|
+        roles.push(role[0])
+      }
+      roles[1..-1]
+    end
 
+    expected_roles = [
```
aaron-lane: This should be defined with `let :expected_roles`.

test/integration/simple_example/controls/forseti.rb
adrienthebo: @marko7460 yeah, let's just switch this to JSON. We don't benefit from using CSV and it adds complexity and cognitive load; let's go with the common convention.

.kitchen.yml
aaron-lane: The tests have been added to the `shared_vpc` suite.

PR #76, opened 2019-03-19
Private Forseti VMs + support for tags + custom VM access_config

### LINKED ISSUE #31: Make forseti server/client private
Some deployments might require forseti server access to be only through private IP. It would be good if the server can be created without `access_config` parameter for these cases. This can be controlled via `private_server` or `private_client` variable

ocervell:
This adds support for private-only Forseti VMs (no public IPs). It also adds support for instance tags and for customizing the `access_config` block on both the client and server VMs.

We add the `client_private` and `server_private` flags and set them as `true` by default (Forseti should be private by default). If those are set to `false`, public IPs will be automatically created for Forseti Server and/or Client VM(s).

We add the `client_tags` and `server_tags` list inputs that allow to customize Forseti client / server VM tags.

We add the `client_access_config` and `server_access_config` that allow to customize networking for Forseti client / server, such as: assigning a PTR record, assigning an existing external IP to Forseti VMs, and whatever is supported by `google_compute_engine`'s `access_config` block.

This fixes #31


ocervell:
Need to stand up a bastion host to make the tests work (bc/ SSH doesn't work on private VMs from the CI network)

PR #79, opened 2019-03-21
Check for errors in startup scripts and test specific version

marko7460:


test/integration/simple_example/controls/client.rb
```diff
@@ -31,6 +32,21 @@
     its('exit_status') { should eq 0 }
   end
 
+  describe command('forseti inventory list') do
```
adrienthebo: This command is duplicated, let's drop it.

test/integration/simple_example/controls/client.rb
```diff
@@ -31,6 +32,21 @@
...
+    its('exit_status') { should eq 0 }
+  end
+
+  describe command('pip show forseti-security|grep Version') do
+    its('exit_status') { should eq 0 }
+
+    let(:formated_stdout) do
```
adrienthebo: This can be expressed as `its('stdout.chomp') { should match "Version: #{forseti_version[1..-1]}" }`

Also, why are we trimming the `forseti_version` string?

test/integration/simple_example/controls/server.rb
```diff
@@ -36,6 +38,17 @@
     it { should exist }
   end
 
+  describe command('pip show forseti-security|grep Version') do
+    its('exit_status') { should eq 0 }
+
+    let(:formated_stdout) do
```
adrienthebo: See previous comment

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -61,10 +59,8 @@ logrotate /etc/logrotate.conf
 # Change the access level of configs/ rules/ and run_forseti.sh
 chmod -R ug+rwx ${forseti_home}/configs ${forseti_home}/rules ${forseti_home}/install/gcp/scripts/run_forseti.sh
 
-# Install tracing libraries
-pip install .[tracing]
```
adrienthebo: Why did we drop this?

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -61,10 +59,8 @@ logrotate /etc/logrotate.conf
...
-
 # Install Forseti
+echo "Installing Forseti"
```
adrienthebo: Does this change provide value? If not, can we omit it?

test/integration/simple_example/controls/client.rb
marko7460: `var.forseti_verison` has `v` infront. Example `v2.13.0`. I didn't trim it in the output

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
marko7460: I don't know why was that there in the first place at all. That line fails causing the whole script to fail. forseti-security DM script doesn't have `pip install .[tracing]`

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
marko7460: It provided a value when looking through serial console output. If startup script ever fails i find this comment valuable in the console logs.

test/integration/simple_example/controls/client.rb
adrienthebo: If that's the case, let's indicate that in code. Using the subscript operator hides this detail, but we can do something like this:

```ruby
let(:forseti_version) { forseti_version.sub(/\Av/, '') }
```

That way the code directly expresses the fact that we're dropping the 'v' prefix.

marko7460:
@aaron-lane The version is already set in variables.tf. The test checks if that version is installed. Specifying `var.forseti_version` in the examples will require that we update that variable every time we bump the `forseti_version`

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
adrienthebo: Ohh, got it. Good catch!

adrienthebo:
@aaron-lane made some good points about the guarantees that we'll get by hard-coding the version in the tests. We're just changing the test coverage so this is non-functional from the perspective of the end user.  I think that we should move forward with the hard-coded assertion and see how it behaves in terms of test coverage.

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
adrienthebo: In terms of functionality this change is negligible, so we can keep it in. However let's minimize changes like this in the future - if there's some subtlety that we need to draw attention let's add logging statements, but we should avoid leaving debug logging in pull requests.

adrienthebo:
We've got a couple of merge conflicts on the PR - the addition of the `forseti_version` output seems to be the culprit. @marko7460 @aaron-lane what do you think of hardcoding the forseti version assertion within the test itself? Was that closer to the desired behavior we wanted to encode in the tests?

aaron-lane:
@adrienthebo yes, the version should be hard-coded within the test. We need a strong assertion about the default behaviour of the module.

outputs.tf
```diff
@@ -66,5 +66,10 @@ output "forseti-server-storage-bucket" {
 
 output "suffix" {
   description = "The random suffix appended to Forseti resources"
-  value       = "${local.random_hash}"
+  value = "${local.random_hash}"
+}
+
+output "forseti-version" {
```
adrienthebo: Since we're hardcoding the forseti version in the tests we no longer need this output.

test/fixtures/shared_vpc/outputs.tf
```diff
@@ -97,4 +97,5 @@ output "forseti-client-service-account" {
 output "forseti-server-service-account" {
   description = "Forseti Server service account"
   value       = "${module.forseti-shared-vpc.forseti-server-service-account}"
-}
\ No newline at end of file
+}
```
adrienthebo: Avoid changing whitespace; it doesn't add value and adds noise to the history.

You can remove this with the following:

```
git checkout master test/fixtures/shared_vpc/outputs.tf
```

This will change `test/fixtures/shared_vpc/outputs.tf` to the version of this file that's on master.

test/integration/simple_example/controls/client.rb
```diff
@@ -31,6 +32,19 @@
     its('exit_status') { should eq 0 }
   end
 
+  describe command('pip show forseti-security|grep Version') do
+    its('exit_status') { should eq 0 }
+    let(:forseti_version_formated) {
```
adrienthebo: We don't need these let helpers anymore; let's just compare `forseti_version` to `subject.stdout`.

examples/shared_vpc/outputs.tf
```diff
@@ -93,3 +93,5 @@ output "forseti-server-service-account" {
   description = "Forseti Server service account"
   value       = "${module.forseti.forseti-server-service-account}"
 }
+
```
adrienthebo: Let's remove these whitespace diffs, they don't provide value.

PR #81, opened 2019-03-21
Specify host for google_sql_user

adrienthebo:
We're getting spurious CI failures when tearing down tests, with a reported error of a missing host. Since we know the IP address of the Forseti server we can specify this field, and in general restricting the connecting host seems like a reasonable configuration.


adrienthebo:
This pull request was meant as a quick fix for some of our CI issues, but making this change has led to more problems. I'm shelving this work until we've handled more pressing issues.

PR #82, opened 2019-03-24
Add real time enforcer firewall rules

adrienthebo:


modules/real_time_enforcer/main.tf
```diff
@@ -118,3 +118,41 @@ resource "google_compute_instance" "main" {
...
 }
+
+#-------------------------#
+# Enforcer Firewall Rules #
+#-------------------------#
+resource "google_compute_firewall" "rt-enforcer-deny-all" {
+  name                    = "forseti-rt-enforcer-deny-all-${local.random_hash}"
+  project                 = "${local.network_project}"
+  network                 = "${var.network}"
+  target_service_accounts = ["${google_service_account.main.email}"]
```
hshin-g: Nit- should we change the service account name to "rt_enforcer" or something along those lines instead of just "main" so it fits with the naming convention with the rest of the modules?

hshin-g: Additionally, we should have a service account section to clearly mark the service account in main.tf
#--------------------------#
# Enforcer Service Account #
#--------------------------#

adrienthebo: Our Terraform standards guide indicates that resources with a single instance should be named `main`. The `forseti-(client|server)` prefixing done in the other modules adds complexity in naming that doesn't provide value, so I've opted to shorten it here. Since the modules are self contained the resource naming is an internal implementation detail and won't leak outside of the module.

I'll add the comment shortly.

PR #97, opened 2019-03-27
Make enforcer role suffix mandatory, include in real_time_enforcer example

adrienthebo:


examples/real_time_enforcer/main.tf
```diff
@@ -56,5 +63,8 @@ module "real_time_enforcer" {
   enforcer_instance_metadata = "${var.instance_metadata}"
   topic                      = "${module.real_time_enforcer_project_sink.topic}"
 
+  enforcer_viewer_role = "organizations/${var.org_id}/roles/${module.real_time_enforcer_roles.forseti-rt-enforcer-viewer-role-id}"
```
aaron-lane: Let's move these interpolated values to outputs of the roles submodule to simplify usage.

adrienthebo: Addressed in  059722a.

PR #100, opened 2019-04-01
Add support for groups settings scanner

adrienthebo:


variables.tf
```diff
@@ -543,6 +543,35 @@ variable "inventory_email_summary_enabled" {
...
+}
+
+variable "groups_settings_period" {
+  description = "the period of max calls to the g suite groups api"
+  default     = "1.1"
+}
+
+variable "groups_settings_disable_polling" {
+  description = "whether to disable polling for the g suite groups api"
+  default     = "false"
```
hshin-g: Nit: Should be "False" to align with the rest of the disable_polling values.

QQ: Is there a reason these settings are separate from the others?

adrienthebo: I was under the impression that YAML required that `false` to be lower cased, but I was incorrect. In addition it looks like I accidentally lowercased all of these entries, I'll correct that.

In terms of file organization, adding a single scanner is easier if the associated variables are clustered together. The reasoning is that we're sharing the definitions for these variables in the root module and the server module so they can be copied between the files. If we have them split up per-section it's harder to copy the grouped elements, so this makes it a bit easier to share the definitions.

Moving forward I think that we should reconsider how we're templating out the Forseti server configuration, as this flattened approach is pretty labor intensive. Generating the configuration using a python script or somesuch might be easier and more maintainable as the responsibilities of the server configuration file grows.

hshin-g: Agreed.

PR #107, opened 2019-04-03
Add licenses to real-time-enforcer policy files

hshin-g:
+ Add licenses to real time enforcer module policy files


modules/real_time_enforcer/files/policy/bigquery/common.rego
```diff
@@ -1,3 +1,18 @@
+# Copyright 2019 The micromanager Authors. All rights reserved.
```
morgante: Who are the micromanager authors?

If we want to reference a set of multiple authors, we need an authors file: https://g3doc.corp.google.com/company/teams/opensource/releasing/authors.md?cl=head

hshin-g: AUTHORS file here:  https://github.com/cleardataeng/micromanager/blob/master/AUTHORS

Where should this file go?

morgante: The submodule root.

hshin-g: @morgante Do we need to include Forseti and CFT as authors as well?

hshin-g: Added the AUTHORS file with micromanager authors.

morgante: Also need `Google LLC` as an author.

PR #112, opened 2019-04-04
Update versioning.rego

hshin-g:
Change versioning.rego policy file


adrienthebo:
The CI failures are spurious, I've manually verified this with a local test-kitchen run.

PR #120, opened 2019-04-05
Add limited retries to real time enforcer services

adrienthebo:
We've observed random but fairly common CI failures with the real-time-enforcer module, specifically in that the enforcer service is not running. Further inspection into one of these failures showed that the `docker pull` failed due to a network error and the service shut down. Network issues are unavoidable so we should be able to handle these situations with some resilience.

This commit updates the opa-server and enforcer systemd units to restart the services upon error, delay 5 seconds between each attempt, and to stop restarting if 5 starts occur within 120 seconds. The delay between each restart allows transient network issues to resolve, and the start limit burst prevents the service from endlessly restarting in case there's an issue that can't be recovered from through restarting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Extending the adopted spec, each change should have a link to its corresponding pull request appended.
 
+<<<<<<< HEAD
 ## [Unreleased]
 
 ### Changed
 
 - Removed Forseti resources from real time enforcer example. [#119]
 
+||||||| merged common ancestors
+=======
+## [Unreleased]
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Extending the adopted spec, each change should have a link to its corresponding pull request appended.
 
+## [Unreleased]
+
+### Fixed
+
+- Real time enforcer containers will restart when encountering errors. [#120]
```
aaron-lane: The link to #120 needs to be added.

adrienthebo: Fixed, amended commit, repushed.

PR #160, opened 2019-05-15
Fix issue #110

bmenasha:
Defines the new `violations_slack_webhook` variable which  configures the slack
webhook for all notifiers. Per notifier slack hooks are unlikely to be useful. The existing per notifier hooks
`iam_policy_violations_slack_webhook` and `kms_violations_slack_webhook` still work and
take precedence over this new variable if both are defined.

### DIFF
--- a/modules/server/templates/configs/forseti_conf_server.yaml.tpl
+++ b/modules/server/templates/configs/forseti_conf_server.yaml.tpl
@@ -288,7 +288,7 @@ notifier:
             - name: slack_webhook
               configuration:
                 data_format: json  # slack only supports json
-                webhook_url: ${IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK}
+                webhook_url: %{ if IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK != "" }${IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK}{ else }${VIOLATIONS_SLACK_WEBHOOK}%{ endif }
 
         - resource: audit_logging_violations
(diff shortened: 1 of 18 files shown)


modules/server/templates/configs/forseti_conf_server.yaml.tpl
```diff
@@ -301,6 +301,10 @@ notifier:
                 data_format: csv
                 # gcs_path should begin with "gs://"
                 gcs_path: gs://${FORSETI_BUCKET}/scanner_violations
+            - name: slack_webhook
+              configuration:
+                data_format: json  # slack only supports json
+                webhook_url: ${VIOLATIONS_SLACK_WEBHOOK}
```
morgante: By default, this will render as

`webhook_url: ""`

Do we know if Forseti correctly handles that case and ignores the webhook entirely? @hshin-g

bmenasha: Forseti has this code to handle the empty url.

    def run(self):
        """Run the slack webhook notifier"""
        if not self.notification_config.get('webhook_url'):
            LOGGER.warn('No url found, not running Slack notifier.')
            return

but you're right. it would make sense to omit the slack notifier so this warning isn't logged. I'll make that change.

morgante:
@bmenasha It looks like a number of unrelated changes got picked up in this PR. Could you rebase on on master (`git rebase -i master`)?

bmenasha:
ugh. i'll just make these edits again in a new fork and submit a new PR :(

PR #168, opened 2019-05-28
Add Slack webhook support.

bmenasha:
Defines the new `violations_slack_webhook` variable which will configure a slack
webhook for all notifiers. The expectation is that per notifier slack hooks are
unlikely to be useful. The existing `iam_policy_violations_slack_webhook` and `kms_violations_slack_webhook` still work and
take precedence over this new variable if both are defined.

### DIFF
--- a/modules/server/templates/configs/forseti_conf_server.yaml.tpl
+++ b/modules/server/templates/configs/forseti_conf_server.yaml.tpl
@@ -285,10 +285,12 @@ notifier:
             # Create an incoming webhook in your organization's Slack setting, located at:
             # https://[your_org].slack.com/apps/manage/custom-integrations
             # Add the provided URL in the configuration below in `webhook_url`.
+            %{ if IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK != "" || VIOLATIONS_SLACK_WEBHOOK != "" }
             - name: slack_webhook
               configuration:
                 data_format: json  # slack only supports json
-                webhook_url: ${IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK}
+                webhook_url: %{ if IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK != "" }${IAM_POLICY_VIOLATIONS_SLACK_WEBHOOK}%{ else }${VIOLATIONS_SLACK_WEBHOOK}%{ endif }
+            %{ endif }
 
         - resource: audit_logging_violations
           should_notify: ${AUDIT_LOGGING_VIOLATIONS_SHOULD_NOTIFY}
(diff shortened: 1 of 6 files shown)


test/integration/simple_example/controls/server.rb
```diff
@@ -324,8 +324,6 @@
             including(
               "resource" => "iam_policy_violations",
               "should_notify" => true,
-              # This extra bit of verification tests the `iam_policy_violations_slack_webhook` variable
```
morgante: @bmenasha Perhaps we could repurpose this test? I'm fine with dropping it though.

modules/server/templates/configs/forseti_conf_server.yaml.tpl
```diff
@@ -301,6 +303,12 @@ notifier:
                 data_format: csv
                 # gcs_path should begin with "gs://"
                 gcs_path: gs://${FORSETI_BUCKET}/scanner_violations
+            %{ if VIOLATIONS_SLACK_WEBHOOK != "" }
```
ocervell: Should we add a `AUDIT_LOGGING_VIOLATIONS_SLACK_WEBHOOK` variable like on line 288 ? Same for all other webhooks

test/integration/simple_example/controls/server.rb
bmenasha: I'm thinking about moving the whole template into it's own sub-module in the hopes of writing more tests against it. but i don't want to do that in this pull request.

modules/server/templates/configs/forseti_conf_server.yaml.tpl
bmenasha: i was hoping to void webhooks for each resource type. seems like an excessive # of variables and i don't see the use case for having resource specific webhooks.

test/integration/simple_example/controls/server.rb
morgante: Got it, feel free to follow up.

bmenasha:
> The example needs to enable a Slack webhook or the tests need to no longer check for its presence.

I thought I did the latter in this change? is it failing?
thanks

morgante:
Yes, it looks like tests are failing: https://concourse.infra.cft.tips/builds/4294

bmenasha:
thanks i see that i missed one. fixed.

dekuhn:
CI/CD is passing now did the test needing to get fix get resolved?

aaron-lane:
@dekuhn yes, it was fixed by https://github.com/forseti-security/terraform-google-forseti/pull/168/commits/98e8b9f8455c6da23e0a1e64e278aa8bea4dddf9.

PR #192, opened 2019-06-25
Set default variable values to support non org installs

umairidris:


umairidris:
From concourse logs
>>>>>>     Destroy failed on instance <simple-example-local>.  Please see .kitchen/logs/simple-example-local.log for more details

Is that related to this PR?

morgante:
I don't think so, but this one might be:

> google_organization_iam_member.org_read.6: Error applying IAM policy for organization "826592752744": Error setting IAM policy for organization "826592752744": googleapi: Error 500: Internal error encountered., backendError

umairidris:
Hmm, those fields should not be affected if org_id is set. Is there any additional info beyond "backendError"? Also is this consistently failing from this PR but not others?

morgante:
Nope, no additional error.

Could you push a new commit to trigger tests rerunning?

umairidris:
Seems to have been a flaky test. Passed on retry.

main.tf
```diff
@@ -19,7 +19,7 @@ resource "random_id" "random_hash_suffix" {
 }
 
 resource "null_resource" "org_id_and_folder_id_are_both_empty" {
-  count = "${var.org_id == "" && var.folder_id == "" ? 1 : 0}"
+  count = "${length(var.composite_root_resources) == 0 && var.org_id == "" && var.folder_id == "" ? 1 : 0}"
```
jeffmccune: @umairidris @morgante We're working on the integration tests for the upcoming 0.12 release and running into a number of the following errors.  Digging into them, we're unsure why this null resource fails when composite_root_resources has items in the list.  Could you help me understand the intent of adding this variable to the list?

```
Error: Error running command 'false': exit status 1. Output:
```

jeffmccune: No rush on this, we traced the Error to another null_resource, `module.server.null_resource.missing_emails`

morgante: Please open a separate issue instead of commenting on merged PRs.

PR #199, opened 2019-07-05
Drop requirement for an exported service account key

rvandegrift:
This will allow the user to use any SDK supported credential for terraform.


rvandegrift:
I've removed `credentials_path` from the test fixtures.  It looks like Concourse CI is failing, but the details require a login/password so I cannot see the issue.

morgante:
CI failures are unrelated.

Issue #148, opened 2019-05-09
Terraform destroy fails on google_sql_user


ocervell:
I ran into an error while running `tf destroy` on a simple Forseti deployment:

```
Error: Error applying plan:

1 error(s) occurred:

* module.forseti-install-simple.module.server.google_sql_user.root (destroy): 1 error(s) occurred:

* google_sql_user.root: Error, failed to deleteuser root in instance forseti-server-db-ce89442d: googleapi: Error 400: Missing parameter: host., required
```

Another `tf destroy` afterwards also gave me some weird errors:

```
11 error(s) occurred:

* module.forseti-install-simple.module.server.output.forseti-server-storage-bucket: variable "server_config" is nil, but no error was reported
* module.forseti-install-simple.module.client.output.forseti-client-vm-public-ip: Resource 'google_compute_instance.forseti-client' does not have attribute 'network_interface.0
.access_config.0.nat_ip' for variable 'google_compute_instance.forseti-client.network_interface.0.access_config.0.nat_ip'
* module.forseti-install-simple.local.random_hash: local.random_hash: variable "random_hash_suffix" is nil, but no error was reported
* module.forseti-install-simple.module.server.output.forseti-server-vm-ip: variable "forseti-server" is nil, but no error was reported
* module.forseti-install-simple.module.client.output.forseti-client-storage-bucket: Resource 'google_storage_bucket.client_config' does not have attribute 'id' for variable 'go
ogle_storage_bucket.client_config.id'
* module.forseti-install-simple.module.server.output.forseti-server-service-account: variable "forseti_server" is nil, but no error was reported

* module.forseti-install-simple.module.client.output.forseti-client-vm-name: Resource 'google_compute_instance.forseti-client' does not have attribute 'name' for variable 'goog
le_compute_instance.forseti-client.name'
* module.forseti-install-simple.module.server.output.forseti-server-vm-name: variable "forseti-server" is nil, but no error was reported
* module.forseti-install-simple.module.server.output.forseti-server-vm-public-ip: variable "forseti-server" is nil, but no error was reported
* module.forseti-install-simple.module.client.output.forseti-client-service-account: Resource 'google_service_account.forseti_client' does not have attribute 'email' for variab
le 'google_service_account.forseti_client.email'
* module.forseti-install-simple.module.client.output.forseti-client-vm-ip: Resource 'google_compute_instance.forseti-client' does not have attribute 'network_interface.0.networ
[...]

PR #207, opened 2019-07-25
Adding parameter to google_sql_user resource

kevensen:
Signed-off-by: Ken Evensen <kdevensen@google.com>

This helps address issue #148.  By specifying the host parameter, the requirements for the parameter being present on destruction are satisfied.

This appears to be the default behavior intended by the documentation: https://cloud.google.com/sql/docs/mysql/create-manage-users.

### DIFF
--- a/modules/server/variables.tf
+++ b/modules/server/variables.tf
@@ -610,6 +610,11 @@ variable "cloudsql_type" {
   default     = "db-n1-standard-1"
 }
 
+variable "cloudsql_user_host" {
+  description = "The host the user can connect from.  Can be an IP address or IP address range. Changing this forces a new resource to be created."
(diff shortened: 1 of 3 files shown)


modules/server/variables.tf
```diff
@@ -610,6 +610,11 @@ variable "cloudsql_type" {
   default     = "db-n1-standard-1"
 }
 
+variable "cloudsql_user_host" {
+  description = "The host the user can connect from.  Can be an IP address or IP address range. Changing this forces a new resource to be created."
```
morgante: nit: spacing

```suggestion
  description = "The host the user can connect from. Can be an IP address or IP address range. Changing this forces a new resource to be created."
```

PR #209, opened 2019-07-26
Add tracing support

ocervell:
To be added to the release for next Forseti release (once https://github.com/forseti-security/forseti-security/pull/2317 is merged into dev)

### DIFF
--- a/modules/server/variables.tf
+++ b/modules/server/variables.tf
@@ -54,6 +54,11 @@ variable "forseti_run_frequency" {
   default     = "0 */2 * * *"
 }
 
+variable "forseti_enable_tracing" {
+  description = "Enable Forseti tracing"
+  default     = "false"
+}
+
 #--------------------------#
 # Forseti config inventory #
 #--------------------------#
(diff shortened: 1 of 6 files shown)


modules/server/variables.tf
```diff
@@ -54,6 +54,11 @@ variable "forseti_run_frequency" {
   default     = "0 */2 * * *"
 }
 
+variable "forseti_enable_tracing" {
+  description = "Enable Forseti tracing"
+  default     = "false"
```
aaron-lane: New boolean variables should be typed based on Terraform 0.12.

```suggestion
  type        = bool
  default     = false
```

variables.tf
```diff
@@ -56,6 +56,11 @@ variable "forseti_run_frequency" {
   default     = "0 */2 * * *"
 }
 
+variable "forseti_enable_tracing" {
+  description = "Enable Forseti tracing"
+  default = "false"
```
aaron-lane:
```suggestion
  type        = bool
  default     = false
```

morgante:
Replaced by #223.

PR #211, opened 2019-07-31
Set inventory email to false and check for missing sendgrid key

umairidris:
Fix https://github.com/forseti-security/forseti-security/issues/3005

### DIFF
--- a/docs/upgrading_to_v3.0.md
+++ b/docs/upgrading_to_v3.0.md
@@ -14,9 +14,9 @@ version intended for Terraform 0.11.x is 2.3.0.
 
 The `inventory_email_summary_enabled` field now has a default value of `false`,
 instead of the previous default of `true`. It also now requires
-`sendgrid_api_key` to be set to a non empty value as without this Forseti will
+`sendgrid_api_key` to be set to a non empty value. Without this Forseti will
 not be able to successfully send emails. See
-[forset-security#3005](https://github.com/forseti-security/forseti-security/issues/3005).
+[forseti-security#3005](https://github.com/forseti-security/forseti-security/issues/3005).
 
 To continue sending inventory summary emails, update a Terraform
 configuration like the following example:
--- a/docs/upgrading_to_v3.0.md
+++ b/docs/upgrading_to_v3.0.md
@@ -1,0 +1,34 @@
+# Upgrading to v3.0
+
+The v3.0 release of the *forseti* module is a backwards incompatible
+release.
+
+## Migration Instructions
+
+This release is meant for use with Terraform 0.12. If you haven't upgraded and
+need a Terraform 0.11.x-compatible version of this module, the last released
+version intended for Terraform 0.11.x is 2.3.0.
+
+
+### Inventory Email Summary
+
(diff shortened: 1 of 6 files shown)


umairidris:
I am still fairly new to terraform so can you clarify where the breaking change is?

- `inventory_email_summary_enabled`: I believe this is not a breaking change as users who have set it will still get the same behaviour. What changes is users who don't set it will now not get the field set by default (this doesn't seem to be breaking though, and when they run terraform apply it should show the changes being made)
- `email_without_sendgrid_api_key`: this null resource (if it works.. see #212) is a breaking change. A config was still valid before if it set inventory_email_summary_enabled to true without sendgrid. So I think I should revert this change.

> Instead, I'd suggest that we simply set inventory_email_summary_enabled to false within the module if sendgrid_api_key is not defined.

Does this violate the "infra as code" nature of the module? For example, suppose `inventory_email_summary_enabled` to true, and then internally we set it to false for the values in the outputted config. That would seem like an unexpected behaviour from the user's perspective.

morgante:
> What changes is users who don't set it will now not get the field set by default (this doesn't seem to be breaking though, and when they run terraform apply it should show the changes being made)

This is the breaking change I was referring to. Specifically, they would see a change in the behavior of the module without changing their config.

That being said, I think it might be worth taking on the work of including a breaking change to make this clearer.

Can you add a note in the CHANGELOG.md and link to a migration doc ([GKE example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/docs/upgrading_to_v3.0.md)) explaining the change?

umairidris:
> > What changes is users who don't set it will now not get the field set by default (this doesn't seem to be breaking though, and when they run terraform apply it should show the changes being made)
> 
> This is the breaking change I was referring to. Specifically, they would see a change in the behavior of the module without changing their config.
> 
> That being said, I think it might be worth taking on the work of including a breaking change to make this clearer.
> 
> Can you add a note in the CHANGELOG.md and link to a migration doc ([GKE example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/docs/upgrading_to_v3.0.md)) explaining the change?
> 
> > email_without_sendgrid_api_key: this null resource (if it works.. see #212) is a breaking change. A config was still valid before if it set inventory_email_summary_enabled to true without sendgrid. So I think I should revert this change.
> 
> I'm fine with keeping this for now, #212 will need to be addressed holistically anyways.

Done.

docs/upgrading_to_v3.0.md
```diff
@@ -0,0 +1,34 @@
+# Upgrading to v3.0
```
morgante: The new version will be v4.0.

umairidris: Done. I thought since v3 was so recent that it would be ok to put it under there, but makes sense to make it v4.

Can you explain why some fields despite being bools are still interpreted as strings? Is it due to templating? Seems a bit confusing and error prone. If we're making a v4 jump I can also convert all string bools to actual bools as well.

morgante: I think that's just legacy, we can ignore for now.

PR #218, opened 2019-07-31
Separate Out CloudSQL component

kevensen:
This is an attempt at sub-modularization.  I wanted to try a single, common resource for starters.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -119,11 +119,10 @@ module "server" {
   server_tags                                         = var.server_tags
   server_access_config                                = var.server_access_config
   server_private                                      = var.server_private
-  cloudsql_region                                     = var.cloudsql_region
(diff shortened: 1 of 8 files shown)


hshin-g:
@kevensen Thanks for submitting this. What is the rationale for separating out CloudSQL as a submodule?

kevensen:
The rationale being that we want to get to a place where if the user wants to go the GKE route, they don't end up with a server vm.  This is the first attempt of what will be a separation of resources that are common to both the GCE install and GKE install.

hshin-g:
Ok thanks for the explanation. Is this something that we want to do for the 2.19 Forseti release, or is this something that can be pushed to 2.20?

kevensen:
Definitely 2.20

kevensen:
This is superseded by #228

PR #226, opened 2019-08-07
Updating tags to support Forseti v2.19.0

kevensen:


kevensen:
I missed the fact that @hshin-g already took care of this in pending PR https://github.com/forseti-security/terraform-google-forseti/pull/225.

Issue #238, opened 2019-08-29
Sync Policy Library from git repo


gkowalski-google:
Add the ability for Forseti to sync the policy library from a git repo, instead of requiring users to manually update the GCS bucket with policy changes.

PR #239, opened 2019-08-30
Support for Policy Library Sync

gkowalski-google:
Updated core module and server module to support the policy library sync. If this option is enabled Terraform will: 1) create an SSH key and copy it to GCS and provide as output, 2) update the startup script to install docker and copy the private SSH key locally, in addition to providing some environment variables needed by Forseti to setup git-sync as a linux service.

Addresses #238.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Simple usage of the module within your own main.tf file is as follows:
 | policy\_library\_home | The local policy library directory | string | `"$USER_HOME/policy-library"` | no |
 | policy\_library\_repository\_url | The git repository containing the policy-library. | string | `""` | no |
 | policy\_library\_sync\_enabled | Sync config validator policy library from private repository | string | `"false"` | no |
+| policy\_library\_sync\_gcs\_directory\_name | The directory name of the GCS folder used for the policy library sync config | string | `"policy_library_sync"` | no |
 | policy\_library\_sync\_git\_sync\_tag | Tag for the git-sync image | string | `"v3.1.2"` | no |
 | policy\_library\_sync\_ssh\_known\_hosts | List of authorized public keys for SSH host of the policy library repository| string | `""` | no |
 | project\_id | Google Project ID that you want Forseti deployed into | string | n/a | yes |
--- a/README.md
+++ b/README.md
@@ -147,6 +147,11 @@ Simple usage of the module within your own main.tf file is as follows:
 | network | The VPC where the Forseti client and server will be created | string | `"default"` | no |
 | network\_project | The project containing the VPC and subnetwork where the Forseti client and server will be created | string | `""` | no |
 | org\_id | GCP Organization ID that Forseti will have purview over | string | `""` | no |
+| policy\_library\_home | The local policy library directory | string | `"$USER_HOME/policy-library"` | no |
+| policy\_library\_repository\_url | The git repository containing the policy-library. | string | `""` | no |
+| policy\_library\_sync\_enabled | Sync config validator policy library from private repository | string | `"false"` | no |
+| policy\_library\_sync\_git\_sync\_tag | Tag for the git-sync image | string | `"v3.1.2"` | no |
+| policy\_library\_sync\_ssh\_known\_hosts | List of authorized public keys for SSH host of the policy library repository| string | `""` | no |
 | project\_id | Google Project ID that you want Forseti deployed into | string | n/a | yes |
 | resource\_enabled | Resource scanner enabled. | string | `"true"` | no |
 | resource\_violations\_should\_notify | Notify for resource violations | string | `"true"` | no |
@@ -187,6 +192,7 @@ Simple usage of the module within your own main.tf file is as follows:
 | forseti-client-vm-ip | Forseti Client VM private IP address |
 | forseti-client-vm-name | Forseti Client VM name |
 | forseti-cloudsql-connection-name | Forseti CloudSQL Connection String |
+| forseti-server-git-public-key-openssh | The public OpenSSH key generated to allow the Forseti Server to clone the policy library repository |
 | forseti-server-service-account | Forseti Server service account |
 | forseti-server-storage-bucket | Forseti Server storage bucket |
 | forseti-server-vm-ip | Forseti Server VM private IP address |
--- a/modules/server/main.tf
+++ b/modules/server/main.tf
@@ -92,8 +92,9 @@ locals {
     }],
 
   }
-  network_interface = local.network_interface_base[var.server_private ? "private" : "public"]
-  missing_emails    = ((var.sendgrid_api_key != "") && (var.forseti_email_sender == "" || var.forseti_email_recipient == "") ? 1 : 0)
+  network_interface                         = local.network_interface_base[var.server_private ? "private" : "public"]
+  missing_emails                            = ((var.sendgrid_api_key != "") && (var.forseti_email_sender == "" || var.forseti_email_recipient == "") ? 1 : 0)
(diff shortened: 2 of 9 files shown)


modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -76,18 +76,47 @@ echo "${forseti_environment}" > /etc/profile.d/forseti_environment.sh | sudo sh
...
+    sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable"
+    sudo apt update
+    apt-cache policy docker-ce
+    sudo apt install -y docker-ce
+  fi
+
+  # Setup local FS
+  # Note: gsutil is using the -n flag so that once the SSH key is copied locally, it is not overwritten for any subsequent runs of terraform
+  sudo mkdir -p /etc/git-secret
+  sudo gsutil cp -n gs://${storage_bucket_name}/policy_library_sync/* /etc/git-secret/
```
gkowalski-google: I considered deleting the private SSH key from GCS after it was successfully copied locally, however there is a [known issue](https://github.com/GoogleCloudPlatform/magic-modules/pull/1759) with the Terraform module that is required to support granting the Forseti service account the necessary file permissions.

modules/server/main.tf
```diff
@@ -425,6 +432,32 @@ resource "google_storage_bucket" "cai_export" {
   depends_on = [null_resource.services-dependency]
 }
 
+resource "tls_private_key" "policy_library_sync_ssh" {
+  count       = "${var.policy_library_sync_enabled ? 1 : 0}"
+  algorithm   = "RSA"
+}
```
gkowalski-google: Generate the SSH key so that Forseti can clone a private repo. The openssh public key is provided as output. I will update the docs to let the user know they will need to add this public key to their git account.

modules/server/main.tf
```diff
@@ -425,6 +432,32 @@ resource "google_storage_bucket" "cai_export" {
...
+  count   = var.policy_library_sync_enabled && var.policy_library_repository_url != "" ? 1 : 0
+  name    = "${local.policy_library_sync_folder}/ssh"
+  content = tls_private_key.policy_library_sync_ssh[0].private_key_pem
+  bucket  = local.server_bucket_name
+
+  depends_on = [
+    google_storage_bucket.server_config,
+    tls_private_key.policy_library_sync_ssh,
+  ]
+}
```
gkowalski-google: Copy the private SSH key to GCS.

modules/server/main.tf
```diff
@@ -425,6 +432,32 @@ resource "google_storage_bucket" "cai_export" {
...
+
+resource "google_storage_bucket_object" "policy_library_sync_ssh_known_hosts" {
+  count   = var.policy_library_sync_enabled && var.policy_library_sync_ssh_known_hosts != "" ? 1 : 0
+  name    = "${local.policy_library_sync_folder}/known_hosts"
+  content = var.policy_library_sync_ssh_known_hosts
+  bucket  = local.server_bucket_name
+
+  depends_on = [google_storage_bucket.server_config]
+}
```
gkowalski-google: If provided, copy the known hosts to GCS to be used by git-sync.

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -76,18 +76,47 @@ echo "${forseti_environment}" > /etc/profile.d/forseti_environment.sh | sudo sh
...
+  gsutil cp -r gs://${storage_bucket_name}/policy-library ${policy_library_home}/ || echo "No policy available, continuing with Forseti installation"
+fi
 
 # Start Forseti service depends on vars defined above.
 bash ./install/gcp/scripts/initialize_forseti_services.sh
 echo "Starting services."
 systemctl start cloudsqlproxy
+if [ "${policy_library_sync_enabled}" == "true" ]; then
+  systemctl start policy-library-sync
+  sleep 5
```
gkowalski-google: Config-validator requires the policy-library to exist before it will successfully startup, need to test this some more.

kevensen: Really?  That's unfortunate.

README.md
```diff
@@ -147,6 +147,11 @@ Simple usage of the module within your own main.tf file is as follows:
 | network | The VPC where the Forseti client and server will be created | string | `"default"` | no |
 | network\_project | The project containing the VPC and subnetwork where the Forseti client and server will be created | string | `""` | no |
 | org\_id | GCP Organization ID that Forseti will have purview over | string | `""` | no |
+| policy\_library\_home | The local policy library directory | string | `"$USER_HOME/policy-library"` | no |
+| policy\_library\_repository\_url | The git repository containing the policy-library. | string | `""` | no |
+| policy\_library\_sync\_enabled | Sync config validator policy library from private repository | string | `"false"` | no |
+| policy\_library\_sync\_git\_sync\_tag | Tag for the git-sync image | string | `"v3.1.2"` | no |
```
hshin-g: Where is this tag image coming from?

modules/server/main.tf
```diff
@@ -92,8 +92,9 @@ locals {
     }],
 
   }
-  network_interface = local.network_interface_base[var.server_private ? "private" : "public"]
-  missing_emails    = ((var.sendgrid_api_key != "") && (var.forseti_email_sender == "" || var.forseti_email_recipient == "") ? 1 : 0)
+  missing_emails             = ((var.sendgrid_api_key != "") && (var.forseti_email_sender == "" || var.forseti_email_recipient == "") ? 1 : 0)
+  network_interface          = local.network_interface_base[var.server_private ? "private" : "public"]
+  policy_library_sync_folder = "policy_library_sync"
```
hshin-g: It might be good to make this a variable instead of a local, since it seems like it is just a string (vs. something with a local hash, for example).

modules/server/main.tf
```diff
@@ -425,6 +432,32 @@ resource "google_storage_bucket" "cai_export" {
   depends_on = [null_resource.services-dependency]
 }
 
+resource "tls_private_key" "policy_library_sync_ssh" {
+  count     = "${var.policy_library_sync_enabled ? 1 : 0}"
```
hshin-g: Is the "${}" necessary?

gkowalski-google: Updated.

modules/server/main.tf
gkowalski-google: Good idea, I have updated this to use a variable instead.

modules/server/outputs.tf
```diff
@@ -14,14 +14,14 @@
...
+  description = "The connection string to the CloudSQL instance"
+  value       = google_sql_database_instance.master.connection_name
 }
 
-output "forseti-server-vm-ip" {
-  description = "Forseti Server VM private IP address"
-  value       = google_compute_instance.forseti-server.network_interface[0].network_ip
+output "forseti-server-git-public-key-openssh" {
+  description = "The public OpenSSH key generated to allow the Forseti Server to clone the policy library repository."
+  value       = tls_private_key.policy_library_sync_ssh[0].public_key_openssh
```
morgante: @gkowalski-google Have you checked that this doesn't raise any errors if policy library sync is turned off?

gkowalski-google: Yes I have tested with the sync enabled and disabled, no issues with the actual infra setup or with the terraform outputs. In the case where it is disabled, this output is not provided.

Issue #212, opened 2019-07-31
null_resource checks don't work in terraform 0.12


umairidris:
null_resource checks like org_id_and_folder_id_are_both_empty don't seem to work anymore in Terraform 0.12. `terraform apply` continues successfully even if none of composite root resources, org id or folder id is set.

jdyke:
Here is what I see when it errors out:

`Error: 2 problems:

- Invalid template interpolation value: Cannot include the given value in a string template: string required.
- Error running command 'echo 'composite_root_resources= org_id= folder_id=' >&2; false': exit status 1. Output: composite_root_resources= org_id= folder_id=`

jdyke:
Hi @umairidris it appears that the variable `org_id` is being set to an empty string but it needs to have a value input from the user. Can you try specifying `org_id` and re-running your plan/apply?

PR #244, opened 2019-09-04
Update variables.tf

jdyke:
removing empty default string from `org_id` variable. If you submit this var as an empty string, `count = length(var.composite_root_resources) == 0 && var.org_id == "" && var.folder_id == "" ? 1 : 0` in main.tf fails on v11 and v12 of terraform.

Users need to specify a value for `org_id`.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -802,7 +802,6 @@ variable "enable_cai_bucket" {
 #--------#
 variable "org_id" {
   description = "GCP Organization ID that Forseti will have purview over"
-  default     = ""
 }
 
 variable "domain" {


jdyke:
This error is represented in #212

morgante:
@jdyke Have you tried setting a `folder_id`? The requirement is that one of org_id or folder_id is set.

jdyke:
`folder_id` nor `org_id` are marked as required in the readme. https://github.com/forseti-security/terraform-google-forseti/blob/master/README.md 

If you want to keep the variables as empty strings they should at least be marked as required similar to `domain` and `project_id` in the `Inputs` table.

`org_id` is marked as required here: https://github.com/forseti-security/terraform-google-forseti/tree/master/examples/install_simple along with `project_id` but not in the main readme.

morgante:
The problem is that in Terraform it's not possible to say that one of two variables must be provided, hence the empty strings. We could add a note to the README saying that one must be provided though, PR welcome.

jdyke:
That’s true. I’ll submit a PR to add a note so it’s more clear. Thanks morgante!

PR #253, opened 2019-09-16
CloudSQL instance created in the same zone as GCE instances

gkowalski-google:
Set the Zone for the CloudSQL instance to the same way that is being done for the client and server. Minor cleanup changes.

Resolves [Forseti issue #3194](https://github.com/forseti-security/forseti-security/issues/3194).

### DIFF
--- a/modules/cloudsql/main.tf
+++ b/modules/cloudsql/main.tf
@@ -21,22 +21,22 @@ locals {
   random_hash     = var.suffix
   cloudsql_name   = "forseti-server-db-${local.random_hash}"
   network_project = var.network_project != "" ? var.network_project : var.project_id
+  cloudsql_zone   = "${var.cloudsql_region}-c"
 }
 
 #------------------------------------#
 # Forseti Private SQL Database Setup #
 #------------------------------------#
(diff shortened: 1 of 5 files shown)


modules/cloudsql/main.tf
```diff
@@ -21,22 +21,22 @@ locals {
   random_hash     = var.suffix
   cloudsql_name   = "forseti-server-db-${local.random_hash}"
   network_project = var.network_project != "" ? var.network_project : var.project_id
+  cloudsql_zone   = "${var.cloudsql_region}-c"
```
gkowalski-google: This is the only functional change. Everything else is some cleanup of variable definitions and updated terraform syntax.

kevensen:
@gkowalski-google Thanks for the PR.  Please correct me if I am wrong, but this doesn't necessarily ensure that the cloudsql instance ends up in the same zone as the client and server, just the same region (which is a good thing by itself).

kevensen:
@gkowalski-google Please disregard my last.  I see that the zone is being hardcoded to `c`.

Issue #236, opened 2019-08-27
More verbose logging for startup scripts


morgante:
In the startup scripts, we should have more verbose logging of what each step is doing so debugging can be done more easily. As suggested [here](https://github.com/forseti-security/terraform-google-forseti/issues/234#issuecomment-525396019).

In particular, we should add a clear message/warning when internet egress failed.

PR #257, opened 2019-09-17
Adding logging for the Forseti server startup script

gkowalski-google:
Adding verbose logging for the startup script to make it easier to troubleshoot any issues. Added a check to attempt to ping google.com to determine if the server VM has a connection to the internet. You can search for "forseti startup" in the Stackdriver logs to see the overall progress of the startup script.

Addresses #236.

### DIFF
--- a/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
+++ b/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
@@ -4,45 +4,57 @@ set -eu
 # Env variables
 USER=ubuntu
 USER_HOME=/home/ubuntu
+INTERNET_CONNECTION="$(ping -q -w1 -c1 google.com &>/dev/null && echo online || echo offline)"
+
+# Log status of internet connection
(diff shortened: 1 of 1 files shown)


modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -4,45 +4,57 @@ set -eu
...
+  sudo mv cloud_sql_proxy.${cloudsql_proxy_arch} /usr/local/bin/cloud_sql_proxy
+  chmod +x /usr/local/bin/cloud_sql_proxy
 fi
 
 # Install Forseti Security.
 cd $USER_HOME
-rm -rf *forseti*
+if [ $INTERNET_CONNECTION == "online" ]; then
+  rm -rf *forseti*
+fi
```
gkowalski-google: I added this check in to not remove the Forseti installation if there is no public internet connection, as was reported from [this Forseti issue](https://github.com/forseti-security/forseti-security/issues/3186#issue-491757730).

PR #264, opened 2019-09-19
Initial commit of consolidated sub-modularization effort

kevensen:
This supersedes the following PR's:

#256 
#258 
#261 
#263 

- [x] rules
- [x] server_iam
- [x] server_gcs
- [x] server_config
- [x] client_iam
- [x] client_gcs
- [x] client_config
- [x] Test import script - Python installer with Forseti 2.18
- [x] Test import script - Python installer with Forseti 2.19
- [x] Test import script - Python installer with Forseti 2.20
- [ ] Test import script - Python installer with Forseti 2.21

PR #270, opened 2019-09-24
Add READMEs to submodules

hshin-g:
Related to #264 

Added READMEs to submodules for user readability and accessibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,14 +68,14 @@ Either the **folder\_id** or **org\_id** input is required to successfully deplo
 | bucket\_cai\_location | GCS CAI storage bucket location | string | `"us-central1"` | no |
 | buckets\_acl\_violations\_should\_notify | Notify for Buckets ACL violations | string | `"true"` | no |
 | cai\_api\_timeout | Timeout in seconds to wait for the exportAssets API to return success. | string | `"3600"` | no |
-| client\_access\_config | Client instance 'access\_config' block | map(any) | `<map>` | no |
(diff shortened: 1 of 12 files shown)


morgante:
@hshin-g @kevensen FYI I would suggest *not* having readmes on "internal" modules unless we expect users to be directly instantiating them. I don't think users would directly instantiate these submodules themselves, would they?

https://www.terraform.io/docs/modules/#standard-module-structure

hshin-g:
I think there is always at least some possibility of users instantiating the submodules themselves, and if there are users who want to contribute to this module, it would help to be able to break down what each of these modules do at the get-go for easier digestion. We can always take out the READMEs if you'd prefer, but I think that having more documentation isn't necessarily a bad thing.

kevensen:
FWIW, I am rolling this back as part of the submodule rollback.  We can discuss this more in the coming days.

kevensen:
After some more thought, I would add that most of the submodules have dependencies on the other submodules, so instantiating them individually is probably an unlikely scenario.

morgante:
While users *can* still instantiate internal modules themselves, exposing them with a README adds them to the public/documented surface of the module. Meaning we have to be careful about changes in the submodules mark versioned changes.

I'd recommend keeping them internal (and not having READMEs) unless there's a compelling reason to encourage users to consume them directly.

morgante:
You can see the impact of this on the registry where submodules without a README are (correctly) marked as "internal": https://registry.terraform.io/modules/terraform-google-modules/forseti/google/3.0.0

hshin-g:
That is a fair point- no READMEs for internal modules. Thanks!

PR #280, opened 2019-09-27
Set the crontab minute to be random

### LINKED ISSUE #278: Forseti cron job is always running at the top of the hour
## Story
The Forseti cron job to run the periodic scan every 2 hours is scheduled to run at the same time for all Terraform installations. With the Python installer, this was random. Running the periodic scan at the same time for all installations is causing some issues for the CAI export and some users might see throttling occur. This would potentially break the inventory process.

## Proposed Solution
Update the cron expression to accept a time to run to be provided by the user, otherwise default to a random time. The cron expression is controlled as a [variable](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/server/variables.tf#L54).

## Acceptance Criteria

- Terraform installations and upgrades have the periodic scan run at a random time.
- Consider a communication to customers in case they see throttling and don't know why.

blueandgold:
This changes the crontab minute bit to be random so that we avoid overloading API services when all the deployed forseti instances start their periodic runs at the same time.

I think the variable will allow users to specify their own crontab string via the CLI or env variable.

Fixes #278.


morgante:
How did you test this? Resource references in variable defaults are unlikely to work (the variable default should be deterministic).

blueandgold:
Thanks for the quick review.  This is my first terraform PR, so I don't know the intricacies of using a `resource` in a `variable`.  I was following the [random_integer example here](https://www.terraform.io/docs/providers/random/r/integer.html).  There are no other examples of how to generate a random number, nor is there a built-in function (that I could see).  Any suggestions?

I tried to test via the cloud shell, but it seems to be cloning from the `master` branch.  Is there any way that I can test this?

morgante:
@blueandgold You can always check out a different branch via git.

modules/server/main.tf
```diff
@@ -94,6 +99,8 @@ locals {
   }
   missing_emails    = ((var.sendgrid_api_key != "") && (var.forseti_email_sender == "" || var.forseti_email_recipient == "") ? 1 : 0)
   network_interface = local.network_interface_base[var.server_private ? "private" : "public"]
+
+  forseti_run_frequency = var.forseti_run_frequency == null ? ${random_integer.random_minute.result} */2 * * *" : var.forseti_run_frequency
```
hshin-g: Missing a quotation here

hshin-g:
@blueandgold After checking out your branch from Git, in your `main.tf` you should change the `source` parameter in your `forseti module` to the path of the local Terraform repo, e.g.:

`source                          = "./terraform-google-forseti"`

blueandgold:
The random minute worked.  Tested it via a deploy to GCP project.
image

modules/server/main.tf
blueandgold: Done.

blueandgold:
Tested this on updating an existing install:
```
Plan: 3 to add, 0 to change, 2 to destroy.

  # module.forseti333.module.client.google_compute_instance.forseti-client must be replaced
-/+ resource "google_compute_instance" "forseti-client" {

  # module.forseti333.module.server.google_compute_instance.forseti-server must be replaced
-/+ resource "google_compute_instance" "forseti-server" {

  # module.forseti333.module.server.random_integer.random_minute will be created
  + resource "random_integer" "random_minute" {
```

PR #284, opened 2019-09-27
[BREAKING CHANGE] Forseti submodules take 2

kevensen:
This is a second attempt at a submodularization effort.  

This PR also includes breaking changes to the "on-gke" submodule.
* Removal of IAM keys as K8s secrets
* Implementation of workload-identity
* Consistent naming of variables/attributes across Terraform and Helm.
* Added the version input variable to the "helm_release" resource.

- [x] Test an import from the Python installer to Terraform state
- [x] Test a deployment with version 4.2.1 of this module and then import to a state compatible with this module.


docs/upgrading_to_v5.0.md
```diff
@@ -0,0 +1,41 @@
+# Upgrading to v5.0
+
+The v4.0 release of the *forseti* module is a backwards incompatible
```
hshin-g: Should be v5.0.

examples/on_gke/main.tf
```diff
@@ -0,0 +1,137 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#--------#
+# Locals #
+#--------#
+
+locals {
+  node_pool_index          = [for index, node_pool in data.google_container_cluster.forseti_cluster.node_pool : index if node_pool.name == var.gke_node_pool_name][0]
```
hshin-g: What is the output for this variable?

kevensen: an integer

modules/cloudsql/main.tf
```diff
@@ -0,0 +1,106 @@
...
+data "google_compute_network" "cloudsql_private_network" {
+  name    = "${var.network}"
+  project = "${local.network_project}"
+}
+
+resource "google_project_service" "service_networking" {
+  count              = var.cloudsql_private ? 1 : 0
+  project            = var.project_id
+  service            = "servicenetworking.googleapis.com"
+  disable_on_destroy = "false"
```
hshin-g: Can this be a bool?

modules/cloudsql/main.tf
```diff
@@ -0,0 +1,106 @@
...
+  prefix_length = 16
+  network       = "${data.google_compute_network.cloudsql_private_network.self_link}"
+  depends_on    = ["google_project_service.service_networking"]
+}
+
+resource "google_service_networking_connection" "private_vpc_connection" {
+  count                   = var.cloudsql_private ? 1 : 0
+  network                 = "${data.google_compute_network.cloudsql_private_network.self_link}"
+  service                 = "servicenetworking.googleapis.com"
+  reserved_peering_ranges = ["${google_compute_global_address.private_ip_address[count.index].name}"]
```
hshin-g: What is the [count.index] used for here?

modules/cloudsql/main.tf
```diff
@@ -0,0 +1,106 @@
...
+# Forseti SQL database #
+#----------------------#
+
+resource "google_sql_database_instance" "master" {
+  name             = local.cloudsql_name
+  project          = var.project_id
+  region           = var.cloudsql_region
+  database_version = "MYSQL_5_7"
+
+  settings {
```
hshin-g: To note, will need to add in a `settings.database_flags` section that has `net_write_timeout = 240` for optimization branch changes.

modules/cloudsql/variables.tf
```diff
@@ -0,0 +1,75 @@
...
+}
+
+variable "cloudsql_private" {
+  description = "Whether to enable private network and not to create public IP for CloudSQL Instance"
+  default     = "false"
+}
+
+variable "cloudsql_type" {
+  description = "CloudSQL Instance size"
+  default     = "db-n1-standard-1"
```
hshin-g: Should change this to `db-n1-standard-4`

modules/cloudsql/main.tf
```diff
@@ -0,0 +1,106 @@
...
+    activation_policy = "ALWAYS"
+    disk_size         = var.cloudsql_disk_size
+
+    backup_configuration {
+      enabled            = true
+      binary_log_enabled = true
+    }
+
+    ip_configuration {
+      ipv4_enabled    = var.cloudsql_private ? "false" : "true"
```
hshin-g: Can this be bool?

modules/on_gke/main.tf
```diff
@@ -14,61 +14,149 @@
...
+  provisioner "local-exec" {
+    command     = "echo 'composite_root_resources=${var.composite_root_resources} org_id=${var.org_id} folder_id=${var.org_id}' >&2; false"
+    interpreter = ["bash", "-c"]
+  }
+}
 
-resource "google_service_account_key" "server_key" {
-  service_account_id = var.forseti_server_service_account
+resource "null_resource" "email_without_sendgrid_api_key" {
+  count = var.inventory_email_summary_enabled == "true" && var.sendgrid_api_key == "" ? 1 : 0
```
hshin-g: Should "true" be bool? Variable here is bool: https://github.com/forseti-security/terraform-google-forseti/blob/1d2a79c3d43b6b285863dcc7595e8f0a1ec14711/modules/on_gke/variables.tf#L667

modules/on_gke/main.tf
```diff
@@ -14,61 +14,149 @@
...
+
+  forseti_run_frequency = var.forseti_run_frequency == null ? "${random_integer.random_minute.result} */2 * * *" : var.forseti_run_frequency
 }
 
 //*****************************************
 //  Allow GKE Service Account to read GCS
 //*****************************************
 
 resource "google_project_iam_member" "cluster_service_account-storage_reader" {
+  count   = var.gke_service_account != "default" ? 1 : 0
```
hshin-g: This var.gke_service_account variable does not seem to have a "default" value that I can find. Is this the intended behavior?

modules/on_gke/main.tf
```diff
@@ -14,61 +14,149 @@
...
 }
 
+#-------------------#
+# Activate services #
+#-------------------#
+resource "google_project_service" "main" {
+  count              = length(local.services_list)
+  project            = var.project_id
+  service            = local.services_list[count.index]
+  disable_on_destroy = "false"
```
hshin-g: Can this be changed to bool?

modules/on_gke/main.tf
```diff
@@ -14,61 +14,149 @@
...
+}
+
 //*****************************************
 //  Obtain Forseti Server Configiration
```
hshin-g: Configiration --> Configuration

main.tf
```diff
@@ -105,52 +103,109 @@ module "client" {
...
+  cloudsql_module      = module.cloudsql
+  server_config_module = module.server_config
+  server_gcs_module    = module.server_gcs
+  server_iam_module    = module.server_iam
+  server_rules_module  = module.server_rules
+
+  services = google_project_service.main.*.service
+}
+
+module "cloudsql" {
```
hshin-g: Does this module also need to include the variable `network`? 

https://github.com/forseti-security/terraform-google-forseti/blob/1d2a79c3d43b6b285863dcc7595e8f0a1ec14711/modules/cloudsql/variables.tf#L20

modules/cloudsql/main.tf
kevensen: This refers to the "google_compute_global_address" resource.  If var.cloudsql_private is false, then neither this nor that google_compute_global_address resource would be deployed.  If it is true, then that address would be added to the list.

Long story short this is a side-effect of using count.

https://github.com/forseti-security/terraform-google-forseti/blob/1d2a79c3d43b6b285863dcc7595e8f0a1ec14711/modules/cloudsql/main.tf#L42-L53

modules/cloudsql/main.tf
kevensen: Per our discussion, we will fix after this PR is merged.

modules/cloudsql/variables.tf
kevensen: ack

modules/on_gke/main.tf
kevensen: Yes.  Interestingly when you create a GKE cluster with the "default" service account, either with Terraform, gcloud or the GCP console, the service account name returned by the GKE data source is "default".

main.tf
kevensen: yes

modules/rules/main.tf
```diff
@@ -61,7 +61,7 @@ resource "google_storage_bucket_object" "main" {
   count   = local.rules_count
   name    = element(local.files, count.index)
   content = element(data.template_file.main.*.rendered, count.index)
-  bucket  = var.bucket
+  bucket  = var.server_gcs_module.forseti-server-storage-bucket
 
   lifecycle {
```
hshin-g: Just curious- do we need to provide the `folder_id` [here ](https://github.com/forseti-security/terraform-google-forseti/blob/c73c874f833afb2759db08e09271707332b5c50f/modules/rules/main.tf#L54)as well if there are users who do not use at the org level?

modules/server_config/outputs.tf
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "forseti-server-config" {
+  description = "The rendered Forseti server configuration file"
+  value       = data.template_file.forseti_server_config.rendered
+}
+
+output "forseti-server-config-md5" {
```
hshin-g: What are these outputs used for?

modules/client_gcs/main.tf
```diff
@@ -0,0 +1,40 @@
...
+}
+
+#------------------------#
+# Forseti storage bucket #
+#------------------------#
+resource "google_storage_bucket" "client_config" {
+  name          = local.client_bucket_name
+  location      = var.storage_bucket_location
+  project       = var.project_id
+  force_destroy = "true"
```
hshin-g: Should this be bool?

modules/client/main.tf
```diff
@@ -24,19 +24,11 @@ locals {
...
-
-  client_project_roles = [
-    "roles/storage.objectViewer",
-    "roles/cloudtrace.agent",
-  ]
+
+  client_conf_path = "${var.forseti_home}/configs/forseti_conf_client.yaml"
+
+  client_name = "forseti-client-vm-${var.suffix}"
+  client_zone = "${var.client_region}-c"
```
hshin-g: In the import script [here](https://github.com/forseti-security/terraform-google-forseti/blob/c73c874f833afb2759db08e09271707332b5c50f/helpers/import.sh#L35), it looks like users are using the entire zone name, which includes the `-c` or whatever they have appended. If the user's import does not have `-c` appended at the end, will this VM be destroyed and then recreated with a zone that has `-c` appended instead?

main.tf
```diff
@@ -89,10 +89,8 @@ module "client" {
 
   project_id               = var.project_id
   client_boot_image        = var.client_boot_image
-  server_address           = module.server.forseti-server-vm-ip
   subnetwork               = var.subnetwork
   forseti_home             = var.forseti_home
-  storage_bucket_location  = var.storage_bucket_location
   forseti_version          = var.forseti_version
```
hshin-g: Should remember to update `forseti_version` here and in the other sub modules.

modules/rules/main.tf
kevensen: Can you please help me understand a little better?  Something along the lines of
```hcl
name = "${var.folder_id}/element(local.files, count.index)"
```

modules/server_config/outputs.tf
kevensen: They are used to prevent the root module from trying to pull down the config for GKE before it gets created in the GCS bucket.

https://github.com/forseti-security/terraform-google-forseti/blob/c73c874f833afb2759db08e09271707332b5c50f/modules/on_gke/main.tf#L104-L118

modules/client/main.tf
kevensen: This is true.  This is actually a problem regardless of submodules.  May need to put another input variable here and in the server.

modules/rules/main.tf
hshin-g: I guess this is more for users who are just using the scanner files out of the box- for example [this](https://github.com/forseti-security/terraform-google-forseti/blob/c73c874f833afb2759db08e09271707332b5c50f/modules/rules/templates/rules/bigquery_rules.yaml) scanner file/template uses the ${org_id} variable that you input- if they do not provide an `org_id` variable, the values would just be `""` and would not necessarily be helpful for users who are only at the folder level, for example, so they would need to probably go in and manually change.

Should be ok though..

modules/server_config/outputs.tf
hshin-g: Got it- thanks!

hshin-g:
Looks good! Added a few comments.

Do we also need to update the examples (e.g. GKE example looks like it's still creating an extra VM)?

modules/rules/main.tf
kevensen: I understand what you are suggestion.  Thanks for the clarification.  I think we do need to add it.

kevensen: Though none of the rule templates as written have a folder_id.  On second thought, let's leave it out for the moment.

kevensen:
> Looks good! Added a few comments.
> 
> Do we also need to update the examples (e.g. GKE example looks like it's still creating an extra VM)?

Yes.  Looks like I need to update the examples per the 0.12 style too.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Extending the adopted spec, each change should have a link to its corresponding pull request appended.
 
+## [v5.0.0] - 2019-09-23
```
hshin-g: Should remember to change this date later on

hshin-g:
@kevensen Quick question- are all of the variables in `server/variables.tf` necessary? It looks like they are all in `server_config/variables.tf` already.

kevensen:
I'll double check first thing in the morning

kevensen:
Yes.  I need to clean this up

PR #292, opened 2019-10-05
Update server config template to reflect CSCC notifier changes

kevensen:
This PR supports setting the rate-limit for the CSCC notifier.  This change has been merged in the forseti-security repo: https://github.com/forseti-security/forseti-security/pull/3258.


modules/server/templates/configs/forseti_conf_server.yaml.tpl
```diff
@@ -272,6 +268,10 @@ scanner:
 ##############################################################################
 
 notifier:
+    api_quota:
+        securitycenter:
+            max_calls: ${SECURITYCENTER_MAX_CALLS}
```
hshin-g: May need to change the `securitycenter_max_calls` variable - on the Forseti PR, it's at 14, but it looks like it is at 1 on this repo: https://github.com/forseti-security/terraform-google-forseti/blob/a3cc0285f2b64976f54a24836a9e4eafa2b32821/modules/server/variables.tf#L227.

kevensen: Thanks for that catch.  The quota is 1000 qpm for CSCC writes and 1000 qpm for CSCC reads.  After some tweaking, I found that 14 qps was safe enough.  I'll update the input variable.

PR #306, opened 2019-10-13
Updating helper scripts to include GKE related roles

kevensen:


examples/on_gke_end_to_end/README.md
```diff
@@ -34,19 +34,21 @@ The [project factory](https://github.com/terraform-google-modules/terraform-goog
 - [Terraform Provider for GCP](https://www.terraform.io/docs/providers/google/index.html) v2.9
 
 ### Configure a Service Account
```
hshin-g: This should have a different number of hashes (i.e. text should be bigger) than the section below it, or else it should be removed if it's not being used.

helpers/setup.sh
```diff
@@ -185,6 +190,18 @@ if [[ -n "$WITH_ENFORCER" ]]; then
   done
 fi
 
+if [[ -n "$ON_GKE" ]]; then
+  gke_roles=("roles/container.admin" "roles/iam.serviceAccountAdmin" "roles/compute.networkAdmin" "roles/resourcemanager.projectIamAdmin")
```
hshin-g: The "roles/iam.serviceAccountAdmin" role looks like it is already added in line 147- does it need to be removed here?

kevensen: We can take it out!

helpers/cleanup.sh
```diff
@@ -41,9 +42,10 @@ ORG_ID=""
 SERVICE_ACCOUNT_NAME=""
 WITH_ENFORCER=""
 HOST_PROJECT_ID=""
+ON_GKE=""
 
 OPTIND=1
-while getopts ":hef:p:o:s:" opt; do
+while getopts ":hekf:p:o:s:" opt; do
```
hshin-g: Slight nit, but would be nice if the getopts order was the same as in `setup.sh`.

In `cleanup.sh`, it is: `":hekf:p:o:s:"`
In `setup.sh`, it is: `":hekp:f:o:"`

kevensen: agreed

examples/on_gke_end_to_end/README.md
kevensen: Good catch.  That should be removed.

Issue #298, opened 2019-10-09
Resolve dependency for Workload Identity in IAM


gkowalski-google:
## Story
During deployment using the latest helm/Terraform changes, there was an error applying a configuration targeting the On GKE End to End example. The workload identity was not available when Terraform was attempting to bind to the service accounts: `Identity namespace does not exist`. This blocked the pods from being deployed onto the cluster. This has been seen with Kube 1.13 and 1.14.

<img width="1387" alt="screenshot" src="https://user-images.githubusercontent.com/10247435/66516138-3f079900-ea95-11e9-869c-8dd376cf8473.png">

## Proposed Solution
Determine if the [service account resources](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/on_gke/main.tf#L123-L139) need to have a dependency on the workload identity.

## Acceptance Criteria
Deploy the GKE End to End example without issue.

gkowalski-google:
This should be resolved with the [latest on GKE end-to-end example](https://github.com/forseti-security/terraform-google-forseti/blob/master/examples/on_gke_end_to_end/main.tf#L86) using GKE TF module 7.2.0. Has not been encountered recently, re-open if necessary.

gkowalski-google:
[This PR](https://github.com/forseti-security/terraform-google-forseti/pull/541) contains a fix for this issue.

PR #307, opened 2019-10-13
Setting the GKE node version in end-to-end example

kevensen:


gkowalski-google:
Addresses #298

examples/on_gke_end_to_end/main.tf
```diff
@@ -132,6 +132,7 @@ module "gke" {
   remove_default_node_pool = true
   identity_namespace       = "${var.project_id}.svc.id.goog"
   node_metadata            = "GKE_METADATA_SERVER"
+  kubernetes_version       = "1.13"
```
hshin-g: Does this need to be exactly 1.13, or can we do something like `~>1.13`?

gkowalski-google: We found an issue with Kube 1.14 related to missing gCloud credentials.

hshin-g: Ah interesting- thanks for catching that!

kevensen: For this, it doesn't accept the Terrafrom version syntax.  What this does is sets the major/minor.  But, it can be any patch release.

https://cloud.google.com/kubernetes-engine/versioning-and-upgrades

PR #316, opened 2019-10-15
Add gcloud command to set project in Cloudshell tutorial

hshin-g:


examples/install_simple/tutorial.md
```diff
@@ -21,6 +21,8 @@ You will need to activate a few APIs on this project for Forseti to function.
 Note: If this step is blocked by an error "Unable to enable required APIs", navigate to the APIs and Services page on the GCP console and manually enable the specified APIs. Alternatively, you can use the following `gcloud` commands:
 
 ```bash
+gcloud config set project PROJECT_ID
```
blueandgold: Can we make this smarter?  I think we can populate this with project id from step 1?  For example, below, you see how it picked up the actual project id?

image

PR #318, opened 2019-10-16
Added sql instance user and password.

joecheuk:
As for KMS, I don't see any direct benefit on using that because if the user can ssh to the VM, they will be able to decrypt the encoded password anyway (due to elevated permission on the VM SA account).

Resolves: forseti-security/forseti-security#3217
Relates: forseti-security/forseti-security#3266

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -850,6 +850,16 @@ variable "cloudsql_net_write_timeout" {
   default     = "240"
 }
 
+variable "cloudsql_db_user" {
+  description = "CloudSQL database user"
+  default     = "forseti_security_user"
+}
+
+variable "cloudsql_db_password" {
+  description = "CloudSQL database password"
+  default     = ""
+}
+
 #----------------#
 # Forseti bucket #
 #----------------#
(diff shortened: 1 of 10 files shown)


variables.tf
```diff
@@ -850,6 +850,16 @@ variable "cloudsql_net_write_timeout" {
...
+variable "cloudsql_db_user" {
+  description = "CloudSQL database user"
+  default     = "forseti_security_user"
+}
+
+variable "cloudsql_db_password" {
+  description = "CloudSQL database password"
+  default     = ""
+}
+
```
kevensen: Can you please add the same variables to [modules/on_gke/variables.tf](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/on_gke/variables.tf)?

joecheuk: Updated, PTAL

kevensen:
@joecheuk, it looks like you have some integration failures.  Just FYI, I updated your branch.

kevensen:
@joecheuk With https://github.com/forseti-security/helm-charts/pull/49 merged, please feel free to add the username and password in the helm charts.

PR #319, opened 2019-10-16
Making the rules_path an input variable

kevensen:


modules/server_config/templates/configs/forseti_conf_server.yaml.tpl
```diff
@@ -219,7 +219,7 @@ scanner:
     # gs://bucket-name/path/for/rules_path
     # if no rules_path is specified, rules are
     # searched in /path/to/forseti_security/rules/
-    rules_path: /home/ubuntu/forseti-security/rules
+    rules_path: gs://${RULES_PATH}
```
hshin-g: Should make this ${RULES_PATH} or clarify that variable "rules_path" in `variables.tf` should not include `gs://` as path of the name.

kevensen: uh-oh.  Yes.

kevensen:
@hshin-g Thanks for the approval.  I will wait to merge until 5.0.0 is released.

PR #320, opened 2019-10-16
Correct server bucket variable being passed to helm_release resource.

kevensen:


modules/on_gke/main.tf
```diff
@@ -248,6 +248,11 @@ resource "helm_release" "forseti-security" {
     value = var.cloud_profiler_enabled
   }
 
+  set {
+    name  = "orchestrator.config.bucket"
```
hshin-g: I don't see this variable in the Helm chart's [variables.tf](https://github.com/forseti-security/helm-charts/blob/master/charts/forseti-security/values.yaml). Is there another place that this would be?

kevensen: Not yet implemented.  Working on a PR that would include this.

FWIW, passing a variable to Helm without a receiving variable in the chart is a no-op.  That said, my desire is to have this in master to support the development of the Helm chart PR.

kevensen: BTW, checking for a corresponding value in the values.yaml was the right thing to do.  Thanks for doing that.

PR #324, opened 2019-10-18
Forseti on-GCE Private Example

### LINKED ISSUE #254: Provide a Complete Configuration for a Private Deployment of Forseti
This purpose of this issue is to capture tasks necessary to provide a working deployment of Forseti in a private configuration.

* Private server and client: #31 and #76 
* Private CloudSQL: #228

Tasks:
Create an example that includes the following:
- [x] CloudNAT
- [ ] Bastion Host
- [ ] Configures master authorized networks for GKE
- [x] Sets all the pertinent variables to private w/o user input: server_private, client_private, cloudsql_private

kevensen:
Resolves #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,7 @@ your environment.
 | forseti-client-storage-bucket | Forseti Client storage bucket |
 | forseti-client-vm-ip | Forseti Client VM private IP address |
 | forseti-client-vm-name | Forseti Client VM name |
+| forseti-client-vm-zone | Forseti Client VM zone |
 | forseti-cloudsql-connection-name | Forseti CloudSQL Connection String |
 | forseti-server-git-public-key-openssh | The public OpenSSH key generated to allow the Forseti Server to clone the policy library repository. |
 | forseti-server-service-account | Forseti Server service account |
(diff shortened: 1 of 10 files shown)


gkowalski-google:
I think this looks good. My only concern is that we now have another example, and I don't think we are testing most of these examples each release. If private is the recommended approach, then can we make the tutorial/install_simple example use private as default? Outside the scope of this ticket, but I also think we should remove some of the other examples.

kevensen:
@gkowalski-google Perhaps we do make the simple_example and/or the install_simple private by default.  This is worth discussing before this PR is finalized.

As far as removing the other examples, we need to be discerning.  The examples highlight relevant deployment scenarios.  I suggest we exercise the different examples in integration testing.

gkowalski-google:
Definitely we should automate testing for each one. How about we just eliminate the simple_example? I think the main difference with this example is that it is making everything private, which is not clear. Since we have your new private example (whether it gets implemented as default example or separately), let's eliminate this one. If there is still a need for it, then let's rename at least.

morgante:
> Definitely we should automate testing for each one. How about we just eliminate the simple_example?

I'd encourage simply removing `simple_example`, as it's mostly redundant with `install_simple` at this point (and change tests to test install_simple).

aaron-lane:
I suggest that the network configuration be included in the example to remove any ambiguity.

PR #326, opened 2019-10-18
Adding manage_server_config_enabled variable 

kevensen:
**Significant Change:** The _manage_server_config_enabled_ variable, when set to false, will disable Terraform from writing a new forseti_conf_server.yaml file if one previously existed in GCS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Version 5.0.0 is a backwards-incompatible release. Please see the [upgrade instr
 - Cleaning up unused input variables. [#300]
 - Add ability to set net_write_timeout for CloudSQL [#299]
 - Update server config template to reflect CSCC notifier changes [#292]
+- Adding manage_server_config_enabled variable [#326]
 
 ## [v4.3.0] - 2019-10-03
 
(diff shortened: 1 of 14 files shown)


helpers/import.sh
```diff
@@ -46,7 +46,7 @@ Example of required Terraform configuration:
 
     module "forseti" {
       source = "terraform-google-modules/forseti/google"
-      version = "~> 4.2"
+      version = "~> 5.0"
```
blueandgold: This is a significant change and should be highlighted in title and description.  So, that it's easy to find when people are just scanning the PR.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Version 5.0.0 is a backwards-incompatible release. Please see the [upgrade instr
 - Cleaning up unused input variables. [#300]
 - Add ability to set net_write_timeout for CloudSQL [#299]
 - Update server config template to reflect CSCC notifier changes [#292]
+- Adding manage_server_config_enabled variable [#326]
```
blueandgold: There is another bucket with states that should be preserved, and that is the `forseti-client-xxxxxx` bucket.

helpers/import.sh
```diff
@@ -62,8 +62,9 @@ Example of required Terraform configuration:
       client_instance_metadata = {
         enable-oslogin = "TRUE"
       }
-      enable_write         = true
-      manage_rules_enabled = false
+      enable_write                 = true
+      manage_rules_enabled         = false
+      manage_server_config_enabled = false
```
blueandgold: So, there is a new change in the `forseti_conf_server.yaml`, in this release.  How will user pick this up if this is disabled during the upgrade?

https://github.com/forseti-security/forseti-security/commit/fdfc1f90e9ca6099040a9d89fb33dc8f07a5f881#diff-c086fc0634ee9757f0ca86cd4a3df527

CHANGELOG.md
kevensen: I didn't add this to the client bucket for the following reason.  Each time the server VM is destroyed and recreated, the server vm gets a new internal IP.  Therefore the config in the client bucket needs to be updated.

helpers/import.sh
kevensen: @blueandgold PTAL above

PR #329, opened 2019-10-18
Module release 5.0.0

gkowalski-google:


paskal:
Trying to follow [upgrade_forseti_with_v5.0/tutorial.md](https://github.com/forseti-security/terraform-google-forseti/blob/041cc03a8773dd012228a64197f3bdbbd18f8aed/examples/upgrade_forseti_with_v5.0/tutorial.md) gives me following (without importing existing state, doing so doesn't help):
```
➜  forseti-deployment ✗ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

module.forseti.module.client.data.template_file.forseti_client_environment: Refreshing state...
module.forseti.module.cloudsql.data.google_compute_network.cloudsql_private_network: Refreshing state...

------------------------------------------------------------------------

Error: Invalid index

  on .terraform/modules/forseti/forseti-security-terraform-google-forseti-1ef45bf/modules/server/outputs.tf line 19, in output "forseti-server-git-public-key-openssh":
  19:   value       = tls_private_key.policy_library_sync_ssh[0].public_key_openssh
    |----------------
    | tls_private_key.policy_library_sync_ssh is empty tuple

The given key does not identify an element in this collection value.
```

paskal:
Anther issue is that import didn't work well for me and tried to re-create everything in server service account (and my deploy account doesn't have rights to do so), it seems that import lost an important part which I returned in #354 with the new naming.

As a side note, asking to deploy Forseti with forseti-server service account is not good at all as permissions are way too big for that. I hope that in the future you will provide separate recommendations on the deployment account permissions.

kevensen:
@paskal THank you for bringing the issue to our attention.  It appears we may have missed a test scenario.  I am about to merge this PR into the master branch.  I will update your PR #354 to point to master.

In that PR, do you mind providing a bit more detail?  For example, how did you deploy Forseti before running the import (TF or Python installer)?  What version did you deploy?

Thank you for the pastebin.  That helps provide context.  Can you add the pastebin to your PR?

PR #339, opened 2019-10-22
Updated main.tf to pass the region to the CloudSQL region.

### LINKED ISSUE #334: Ensure Server VM & CloudSQL are Deployed Together
For a more performant inventory, the Server VM and the CloudSQL should be deployed together in the same region.

But, in the Simple Install, the `region` value in terraform.tfvars would allow the server vm to be deployed to a different region (either as the default or as user-specified) from the cloudsql region (which is us-centra1).

Essentially, the `terraform.tfvars` of simple install needs to be updated with:
```
cloudsql_region = "us-central1"
server_region   = "us-central1"
client_region   = "us-central1"
```

image

gkowalski-google:
The install simple example will now deploy the GCE Instance VMs and CloudSQL instances in the same region. I performed a test using us-west2 and verified these resources all got deployed there.

Fixes #334

### DIFF
--- a/examples/install_simple/main.tf
+++ b/examples/install_simple/main.tf
@@ -15,29 +15,30 @@
  */
 
 module "forseti-install-simple" {
-  source  = "terraform-google-modules/forseti/google"
-  version = "~> 5.0.0"
+  source                   = "terraform-google-modules/forseti/google"
+  version                  = "~> 5.0.0"
 
-  project_id = var.project_id
-  org_id     = var.org_id
-  domain     = var.domain
+  project_id               = var.project_id
+  org_id                   = var.org_id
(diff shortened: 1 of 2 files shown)


examples/install_simple/main.tf
```diff
@@ -15,29 +15,30 @@
...
+  org_id                   = var.org_id
+  domain                   = var.domain
 
-  server_region = var.region
-  client_region = var.region
-  network       = var.network
-  subnetwork    = var.subnetwork
+  server_region            = var.region
+  client_region            = var.region
+  cloudsql_region          = var.region
```
gkowalski-google: Only functional change is this line. The rest looks changed because I indented the values to be on the same column.

PR #354, opened 2019-10-25
Improve existing resource import in v5.0.0

paskal:
I'm upgrading terraform from v4.3.0 to v5.0.0, my [main.tf](https://pastebin.com/R8JkEXhW) is by the link.

Current `import.sh` didn't work well for me and tried to re-create every organization-level right in server service account (and my deploy account doesn't have rights to do so). Reason for that is that `import.sh`  lost org rights import [at some point of time](https://github.com/forseti-security/terraform-google-forseti/commit/9ca48ffef91c68c77d8a4999c95b24ec7621e894#diff-fe1cddf0312a004f99e92d20b3ee6232) and this PR re-introduce them.

Also it fixes  `google_project_service.main` keys order: currently they are imported in incorrect order, which triggers their adjustments when in reality it is not necessary.


kevensen:
Hi @paskal .  Just for our information, did you deploy v4.3.0 with the Python installer?  Or was it deployed with Terraform?

paskal:
I've deployed it with Python installer prior to ~v4.1.0 (Forseti 0.22.0), then updated to v4.3.0 (0.22.0), then now updated to v5.0.0 (0.23.0). I've had numerous problems with migration to terraform in the first place because of the inconsistent naming of resources prior to this, but I think it was my own fault.

kevensen:
@paskal Could you please provide the list of permissions your deployer service account does have?  This information will help with our testing.

paskal:
Sure, here is it:
```
// project
 Owner
Storage Admin
Storage Object Viewer
// organization
Security Reviewer 
```

paskal:
The changes I observe before fixing the script:
```diff
  # module.forseti.google_project_service.main[12] must be replaced
-/+ resource "google_project_service" "main" {
      + disable_on_destroy = false
      ~ id                 = "project_id/serviceusage.googleapis.com" -> (known after apply)
        project            = "project_id"
      ~ service            = "serviceusage.googleapis.com" -> "logging.googleapis.com" # forces replacement
    }

  # module.forseti.google_project_service.main[13] must be replaced
-/+ resource "google_project_service" "main" {
      + disable_on_destroy = false
      ~ id                 = "project_id/logging.googleapis.com" -> (known after apply)
        project            = "project_id"
      ~ service            = "logging.googleapis.com" -> "cloudasset.googleapis.com" # forces replacement
    }

  # module.forseti.google_project_service.main[14] must be replaced
-/+ resource "google_project_service" "main" {
      + disable_on_destroy = false
      ~ id                 = "project_id/cloudasset.googleapis.com" -> (known after apply)
        project            = "project_id"
      ~ service            = "cloudasset.googleapis.com" -> "storage-api.googleapis.com" # forces replacement
    }
```
+ re-assigning all rights to server service account.

paskal:
To show what this change produce, I've removed `terraform.tfstate` from my existing v5.0.0 installation and run following:
```
// current master branch
$ curl --location --remote-name https://raw.githubusercontent.com/forseti-security/terraform-google-forseti/7c73c5bc790115d085c79423aece2d88685d8437/helpers/import.sh
$ chmod +x import.sh
$ import.sh <params>
$  terraform plan | grep Plan:
Plan: 36 to add, 4 to change, 9 to destroy.
```
Then I did the same with this PR version of the script:
```
// this PR version
$ curl --location --remote-name https://raw.githubusercontent.com/forseti-security/terraform-google-forseti/7c73c5bc790115d085c79423aece2d88685d8437/helpers/import.sh
$ chmod +x import.sh
$ import.sh <params>
$ terraform plan | grep Plan:
Plan: 17 to add, 4 to change, 1 to destroy.
```

paskal:
Is there anything else I need to clarify?

kevensen:
@paskal Thank you for working on this.  The only issue I see is that if the permissions do not exist and you attempt to import them, the script will display a lot of confusing errors.  Can we handle said errors a little more proactively?

Also, could you please re-base your changes on top of the master branch?

paskal:
This script supposed to be run on top of existing v4.0.0 installation to migrate it to v5.0.0, I think you supposed to have server service account by this point and you want to have it properly imported (which is exactly the problem I had with this script and which I fixed in this PR).
I couldn't think of any possible way not to produce errors and at the same time have some output of the script for debugging the issue: you either have all output or none in this situation. In scenario I found this problem is you'll have no error messages, and whatever you are concerned about is the scenario I'm unaware of.
From my POV it's okay to have meaningful error messages and would be worse hiding them.

paskal:
For reference, these lines already were in the code and were fixed for new format in https://github.com/kevensen/terraform-google-forseti/commit/d8e7651ef13a5631883aa80eec7c3c551f36bc16#diff-fe1cddf0312a004f99e92d20b3ee6232R170-R188 but then removed in https://github.com/kevensen/terraform-google-forseti/commit/692b211de8654f215d2778e2d4834c56a5c03fec#diff-fe1cddf0312a004f99e92d20b3ee6232L174-L188.

kevensen:
@paskal I was incorrectly recalling an error the resulted in us removing those lines.  FWIW, the import script is there to assist with 2 cases.

1. Upgrading Forseti deployed with terraform-google-forseti 4.x
2. Migrating Forseti deployed with the Python installer to be supported by terraform-google-forseti.

I wanted to make sure that the latter case is covered as well.  I've tested installing Forseti v2.18.0 with the Python installer then running your version of the import.sh script.  It looks good.

Issue #341, opened 2019-10-22
Make the Options for Policy Library Sync to Be Consistent


dekuhn:
@blueandgold commented on [Sat Oct 19 2019](https://github.com/forseti-security/policy-library/issues/190)

On GCE, this is off by default, and needs `policy_library_sync_enabled = true` to be set.

On GKE, this is on by default, and `policy_library_sync_enabled` is not exposed to be configure-able.

Since there is a case for continued usage of GCS buckets, a couple of things need to be done:
1: `policy_library_sync_enabled` needs to be configure-able on GKE in case GCS is prefered
2: make consistent the default behavior whether it's on or off, on both GCE and GKE

This needs to be implemented with the corresponding [Helm issue](https://github.com/forseti-security/helm-charts/issues/38).

dekuhn:
We will disable this feature on GKE by default in 2.23 release. If someone wants to use CV + Policy Library they can explicating enable this feature.

GCS sync support on GKE needs to be added. Our default out of the box behavior on GCE/GKE should be GCS Sync.

kevensen:
Closed via #356

PR #356, opened 2019-10-25
Adding GCS as a policy-library store for on-GKE

kevensen:
1. Removed the default **policy_library_repository_url**
2. Added policy_library_sync_enabled for on_gke examples
3. Added Workload Identity for config-validator
4. Corrected some Helm variables per the config-validator chart
5. on-GCE generates a private SSH key.  on-GKE also behaves the same if a private ssh key file is not provided.

Tests:
- [x] on_gke_end_to_end - config-validator disabled && git-sync disabled --> forseti deploys and config validator does not, no keys are created
- [x] on_gke_end_to_end - config-validator disabled && git-sync enabled --> forseti deploys and config validator does not, no keys are created
- [x] on_gke_end_to_end - config-validator enabled && git-sync disabled --> forseti deploys && config validator deploys && policy is pulled from GCS && no keys created
- [x] on_gke_end_to_end - config-validator enabled && git-sync disabled && policy_library_url --> forseti deploys && config validator deploys && policy is pulled from Git && no keys created
- [x] on_gke_end_to_end - config-validator enabled && git-sync enabled -->  forseti deploys && config validator deploys && SSH key is generated
- [x] on_gke_end_to_end - config-validator enabled && git-sync enabled && SSH key filename is provided-->  forseti deploys && config validator deploys && public ssh-key output matches existing public key.
- [x] on_gke_end_to_end [5.0.0 --> 5.1.0] - With config-validator enabled (and git-sync) and upgrade to 5.1.0(?) maintains functionality.  This is the periodic config-validator.


kevensen:
Addresses #341

kevensen:
Need to test
- on_gke example with git-sync enabled.  Then update policies and ensure pod deletes and restarts.

Issue #361, opened 2019-10-29
[v5.0.0] tls_private_key.policy_library_sync_ssh is empty tuple


paskal:
Terraform module v5.0.0 + Forseti v0.23.0 on GCE, upgraded from v4.3.0 + v.0.22.0 Terraform on GCE.

```
➜ terraform --version
Terraform v0.12.12
+ provider.google v2.11.0
+ provider.null v2.1.2
+ provider.random v2.2.1
+ provider.template v2.1.2
+ provider.tls v2.1.1
```

`terraform plan`, `terraform apply` doesn't work for me. Steps to reproduce:

1. Create `main.tf` with following context, filling real `domain`, `project_id`, `org_id`:

```terraform
module "forseti" {
  source = "terraform-google-modules/forseti/google"
  version = "~> 5.0.0"

  # Replace these argument values with those obtained in the Prerequisites section
  domain               = "example.com"
  project_id           = "project_id"
  org_id               = "123456789000"

  client_instance_metadata = {
    enable-oslogin = "TRUE"
  }
  enable_write         = true
  manage_rules_enabled = false
}
```

2. Issue `terraform init`

3. Issue `terraform plan`, my result is:

```
➜  forseti-deployment git:(master) ✗ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

module.forseti.module.client.data.template_file.forseti_client_environment: Refreshing state...
module.forseti.module.cloudsql.data.google_compute_network.cloudsql_private_network: Refreshing state...

------------------------------------------------------------------------

Error: Invalid index

  on .terraform/modules/forseti/forseti-security-terraform-google-forseti-1ef45bf/modules/server/outputs.tf line 19, in output "forseti-server-git-public-key-openssh":
  19:   value       = tls_private_key.policy_library_sync_ssh[0].public_key_openssh
    |----------------
    | tls_private_key.policy_library_sync_ssh is empty tuple

The given key does not identify an element in this collection value.
```

Unless I delete following blocks from forseti terraform module files, `terraform plan` and `terraform apply` don't work for me:

```
➜ fgrep forseti-server-git-public-key-openssh .terraform -r
.terraform/modules/forseti/forseti-security-terraform-google-forseti-1ef45bf/outputs.tf:output "forseti-server-git-public-key-openssh" {
[...]

PR #362, opened 2019-10-29
Support the case where the user is syncing from GCS

kevensen:
With Terraform version `0.12.12`, if a user sets
```
config_validator = true
policy_library_sync_enabled  = false
```
They will encounter an error in the output the public OpenSSH key not being generated.


kevensen:
Fixes #361

dekuhn:
@kevensen @gkowalski-google is this fix assumed targeted for the next release?

kevensen:
@dekuhn Yes

PR #369, opened 2019-11-01
Update GKE end-to-end example for machine type and disk type

### LINKED ISSUE #368: Update GKE configuration to match GCE machine and disk type
## Story
Recently the GCE Server VM was updated to use a bigger machine type and a better disk type. Update Terraform for GKE to have these same changes.

## Proposed Solution
Since the Forseti GKE module just handles the Forseti specifics, users need to configure their Kubernetes cluster/nodes themselves. We have an end-to-end example showing a recommended Kubernetes cluster configuraiton. Update this example to specify the same defaults that GCE uses.

## Acceptance Criteria
GIVEN: A user uses the GKE end-to-end example
WHEN: They apply the Terraform configuration
THEN: The cluster node pool is created with the recommended options.

gkowalski-google:
Updated the On GKE end-to-end example to have the same machine type and disk type as the defaults for GCE. Moved these variables out of main.tf into the variables.tf file.

Fixes #368

### DIFF
--- a/examples/on_gke_end_to_end/main.tf
+++ b/examples/on_gke_end_to_end/main.tf
@@ -118,11 +118,11 @@ module "gke" {
 
   node_pools = [{
     name               = "default-node-pool"
-    machine_type       = "n1-standard-2"
+    machine_type       = var.default_node_pool_machine_type
     min_count          = 1
     max_count          = 1
-    disk_size_gb       = 100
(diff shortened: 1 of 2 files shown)


gkowalski-google:
gke-node-pool

gkowalski-google:
@kevensen Can you please merge this in?

Issue #153, opened 2019-05-14
Ignore changes to cloudsql disk size


Ian-Harris:
The cloudsql instance which is created in module.forseti.module.server.google_sql_database_instance.master has disk_autoresize: "true".  This is all well and good.  However this means that future `terraform apply`s attempt to delete and recreate the database in order to return the disk size to the original 25GB.  This is both i) quite destructive and ii) doomed to failure because of GCP's soft deletion rules.

Would it be possible for a suitable ignore_changes entry to be added to this resource's lifecycle?

hshin-g:
@Ian-Harris there is a pull request that was opened by you and subsequently closed afterwards. Would you be able to address the comments in that PR and we can get this merged in?

Issue #363, opened 2019-10-29
[v5.0.0] CloudSQL unnecessary recreation in case of automatic size increase


paskal:
Hello team!

I have v5.0.0 Terraform module installation on GCE, and found out that in scenario where CloudSQL outgrown default `25`Gb of size (`cloudsql_disk_size` variable), Terraform want to re-create it with smaller disk, see output of `terraform plan` (only relevant part is present):
```bash
Terraform will perform the following actions:

  # module.forseti.module.cloudsql.google_sql_database_instance.master must be replaced
-/+ resource "google_sql_database_instance" "master" {
      ~ settings {
          ~ disk_size                   = 76 -> 25 # forces replacement
     }
}
```
Re-creation would be unsuccessful, see #271 for reason.

Expected behavior: if database size is bigger than value in settings, no changes applied to it, e.g. currently it looks like `disk_size = 25`, but should be `disk_size >= 25` (it's pseudo-code, I'm not familiar with Terraform syntax enough to say how it will look like in reality).

paskal:
Closing in favor of #153.

PR #371, opened 2019-11-04
Ignoring the size of the CloudSQL disk on re-apply

kevensen:
If a CloudSQL disk is auto-resized as the database grows, it will likely not match the parameter in the Terraform module.  On Terraform re-apply, Terraform will attempt to destroy and recreate the CloudSQL database.

This addresses #153.


kevensen:
#363

hshin-g:
Thanks for pushing this! If someone did want to change the size of the CloudSQL disk, would they need to edit out this part and rerun?

kevensen:
@hshin-g Thanks for the review.  Hypothetically, a user shouldn't have to.  By default, the database disk will auto-resize as their data grows.

PR #391, opened 2019-11-16
Configure firewall rules in support of private Client and Server.

### LINKED ISSUE #380: Feature - Disable the deployment of allow-all firewall rules if Forseti is private
## Story
By default, Forseti deploys in a public configuration.  This means the Server, Client and CloudSQL instances get public IP's.  In order to access the Client and Server, a user needs SSH access.  Forseti creates firewalls allow SSH from anywhere (0.0.0.0/0).  If Forseti is deployed in a private configuration, the Client and Server are not given pulbic IP's.  However, the anywhere firewall rules are still deployed.

## Proposed Solution
When the **server** VM is set to private
- [x] Do not add the allow SSH from anywhere rule for the server
- [x] Add a firewall rule to allow SSH from the IAP proxy address space (35.235.240.0/20) to the server with a priority of 100 

When the **client** VM is set to private
- [x] Do not add the allow SSH from anywhere rule for the client.
- [x] Add a firewall rule to allow SSH from the IAP proxy address space (35.235.240.0/20) to the client with a priority of 100 

## Acceptance Criteria
GIVEN: A user wanting to SSH to the server
WHEN: server_private=true
THEN: The user can still SSH to the server without further configuration
GIVEN: A user wanting to SSH to the client
WHEN: client_private=true
THEN: The user can still SSH to the client without further configuration

kevensen:
Resolves #380

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -141,7 +141,7 @@ module "server" {
   cloud_profiler_enabled   = var.cloud_profiler_enabled
   mailjet_enabled          = var.mailjet_enabled
   network                  = var.network
-  network_project          = var.network_project
+  network_project          = local.network_project
   server_grpc_allow_ranges = var.server_grpc_allow_ranges
   server_instance_metadata = var.server_instance_metadata
   server_ssh_allow_ranges  = var.server_ssh_allow_ranges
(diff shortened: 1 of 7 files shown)


kevensen:
Lint will pass after we merge modulerelease510 to master then master into this feature branch.

PR #392, opened 2019-11-18
Add missing space.

xingao267:
Missing a space in the yaml. If users just comment out the whole section and use the rule, then it will fail.


kevensen:
@xingao267 Thanks for this PR.  Can you please update your branch?  When the tests pass, I'll be happy to approve and merge.

Issue #366, opened 2019-10-30
Forseti on-GCE: Specify branch in policy-library


kevensen:
As a user of Forseti on-GCE, I would like to specify a branch other than _master_ for my policy-library. Add `policy_library_repository_branch` to the Forseti readme to list it as an input. This is supported on GKE.

PR #394, opened 2019-11-18
Add policy_library_repository_branch to GCE module

hshin-g:
Addresses #366


hshin-g:
@gkowalski-google No, does there need to be if it's specific to the Terraform module?

gkowalski-google:
> @gkowalski-google No, does there need to be if it's specific to the Terraform module?

Ticket not necessary, but the [initialize forseti services script](https://github.com/forseti-security/forseti-security/blob/master/install/gcp/scripts/initialize_forseti_services.sh#L56) will need to use this variable. Please default to master if none provided.

kevensen:
@hshin-g Do you mind updating your branch?  It should clear up the lint and integration tests.

hshin-g:
Ok that makes sense- thanks for the heads up @gkowalski-google! Will submit a PR for that as well.

Have merged in master- thanks Ken!

PR #400, opened 2019-12-02
Conditional fw rules

bcadiot:
This PR make the creation of firewall rules conditional. This is useful and needed when you deploy forseti in a VPC where you don't have permissions other than juste consuming IPs (Like in most enterprise shared VPC).
Another use case is simply allowing users to create their own custom FW rules.

Note : Another PR is in preparation for doing the same thing with service networking (when in using CloudSQL Privately). The global goal is to be able to deploy forseti in a shared VPC (on GKE or not) when you can only use a subnet and nothing more.

PR #401, opened 2019-12-02
Conditonal service networking

bcadiot:
This PR make the creation of the Service Networking Connection optional and conditional on a boolean. The use case is strictly the same of https://github.com/forseti-security/terraform-google-forseti/pull/400

When you want to use CloudSQL privately and using a Shared VPC, you haven't automatically permissions to create such things. Moreover sometime the service networking connection already exist and in the VPC and you don't want to create another.


kevensen:
Hi @bcadiot.   Thanks for the PR.  Per the [pre-requisites](https://github.com/forseti-security/terraform-google-forseti#service-account), the CFT deployer service account requires `roles/compute.networkAdmin` in order to complete the installation.  That said, let me think about this use-case.

PR #414, opened 2019-12-12
Allow user to configure Scanner Rules path to GCS or local dir

mcapts:
This PR will allow a Terraform Forseti user to provide a rules_path as input to the top-level forseti module, and have that propagate down into the server_config module. We have tested to ensure this works with a GCS path only (not a local dir, although that is the default behavior so this should not affect that). @blueandgold

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -320,6 +320,7 @@ module "server_config" {
   violations_slack_webhook                            = var.violations_slack_webhook
   cscc_violations_enabled                             = var.cscc_violations_enabled
   cscc_source_id                                      = var.cscc_source_id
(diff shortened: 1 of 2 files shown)


kevensen:
Hi @mcapts Thanks for this PR.  This is a useful addition.  Are you able to sign the CLA?

mcapts:
@googlebot I signed it!

blueandgold:
@googlebot rescan

blueandgold:
@googlebot rescan

gkowalski-google:
@mcapts Hello, sorry we have not merged this yet. We are fixing some issues with the build and then will get this merged in.

PR #425, opened 2019-12-17
Increased open files limit

red2k18:
This PR fixes `OSError: [Errno 24] Too many open files` error reported in [this issue](https://github.com/forseti-security/forseti-security/issues/3425).

### DIFF
--- a/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
+++ b/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
@@ -154,6 +154,10 @@ echo "Forseti Startup - Enabling and starting Forseti service."
 systemctl enable --now forseti
 echo "Forseti Startup - Success! The Forseti API server has been enabled and started."
 
+# Increase Open File Limit
+echo "ubuntu soft nofile 32768" | sudo tee -a /etc/security/limits.conf
+echo "ubuntu hard nofile 32768" | sudo tee -a /etc/security/limits.conf
+
 # Create a Forseti env script
 FORSETI_ENV="$(cat << EOF
 #!/bin/bash
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @morgante @aaron-lane @adrienthebo @hshin-g @kevensen
+* @morgante @aaron-lane @adrienthebo @hshin-g @kevensen @gkowalski-google
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -24,6 +24,47 @@ verifier:
   fail_fast: false
 
 suites:
+  - name: install_simple
+    lifecycle:
+      pre_verify:
+        - chmod go-rwx test/fixtures/bastion/tls_private_key_install_simple
+    driver:
+      name: terraform
+      root_module_directory: test/fixtures/install_simple
+      command_timeout: 1800
+    verifier:
+      systems:
+        - name: gcp
+          backend: gcp
+          controls:
+            - forseti
+        - name: server
+          backend: ssh
+          controls:
+            - server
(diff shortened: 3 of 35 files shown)


modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -154,6 +154,10 @@ echo "Forseti Startup - Enabling and starting Forseti service."
 systemctl enable --now forseti
 echo "Forseti Startup - Success! The Forseti API server has been enabled and started."
 
+# Increase Open File Limit
+echo "ubuntu soft nofile 32768" | sudo tee -a /etc/security/limits.conf
```
gkowalski-google: I think it would be good to have this limit as a Terraform variable. It's possible that some users will want/need to set this to a different value and will be easier to maintain/change this in the future as a variable.

gkowalski-google:
The build is failing because we are hitting the limit of IAM members. We will need someone to clean up and re-run the tests.

```
Error: Error applying IAM policy for organization "943740911108": Error setting IAM policy for organization "943740911108": googleapi: Error 400: The number of members in the policy (1,501) is larger than the maximum allowed size 1,500., badRequest
```

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
red2k18: That is a nice to have feature. Setting it to 32768 fixes the `open files error`. We can open an issue to make ulimit modifiable for later.

blueandgold: I agree that it's a great idea to be in terraform, but for simplicity purpose, would like to have a real use case before adding more to terraform.  So, okay to add this here for now.

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -154,6 +154,10 @@ echo "Forseti Startup - Enabling and starting Forseti service."
...
+echo "ubuntu hard nofile 32768" | sudo tee -a /etc/security/limits.conf
```
blueandgold: If someone resets the VM, will this be added to the `limits.conf` file again?

red2k18: Have verified this on `red2k18-terraform-test project`. 
1. Upon resetting the server VM, the new ulimit is applied. 
2. Upon manually removing the code in this PR from the server VM and resetting it, it stays and doesn't reset to 1024 as there is no configuration that overrides 32768. 
3. To change the ulimit to a different value, user needs to edit the ulimit configuration in the startup script of the server VM and reset the VM.

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
red2k18: Opened [an issue](https://github.com/forseti-security/terraform-google-forseti/issues/429) to track this and make ulimit configurable.

modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
red2k18: Just wanted to add that I have tested on `rehana-test` project as well. New ulimit code is picked up after VM reset. Also, inventory is successfully created after modifying the startup script.

red2k18:
Closing this as I opened [a new PR](https://github.com/forseti-security/terraform-google-forseti/pull/450) to get the fix for the errors caused due to upgrading Terraform to 0.12.6.

PR #434, opened 2020-01-07
Set Google Provider version to 2.11.0

### LINKED ISSUE #433: Integration tests are failing to enable Bigquery API
## Story
The integration test build is failing to enable Bigquery API and preventing PRs from being tested/merged.

```
    Error: googleapi: Error 503: The service(s) ["bigquery-json.googleapis.com"] are still being enabled for project ci-forseti-serv-owgl. This isn't a real API error, this is just eventual consistency.
       
         on ../../../main.tf line 77, in resource "google_project_service" "main":
         77: resource "google_project_service" "main" {
```

## Proposed Solution
After some more testing, I found that enabling the `bigquery-json` API fails with older versions of the Google provider (eg. 2.11 and 2.12), however it works with the latest version. As mentioned in [this issue](https://github.com/terraform-providers/terraform-provider-google/issues/4590), version 2.17 resolved this issue with this API and supports both `bigquery` and `bigquery-json`. We have another issue to deal with the version upgrades, and for now we just want to fix the build. In order to do that, I am keeping the same Google Provider version and updating everything to use `bigquery` which is the "new" name for the API.

## Acceptance Criteria
Integration test build passes.

gkowalski-google:
Set the version for the Google provider to 2.11.0. The current syntax is using Google Provider version >= 2.11.0 and < 3.0. This is causing the Provider version to be 2.20.1 which was released on December 13 2019.

Resolves #433

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -18,7 +18,7 @@
 terraform {
   required_version = ">= 0.12"
   required_providers {
-    google   = "~> 2.11"
+    google   = "2.11.0"
     null     = "~> 2.0"
     template = "~> 2.0"
     random   = "~> 2.0"


hshin-g:
Thanks for submitting this @gkowalski-google . Do we also need to update the examples as well?

gkowalski-google:
> I'd strongly discourage pinning to a single provider version here as it will prevent anyone from using other versions entirely and might conflict with common configs.
> 
> The better fix would be to resolve the actual issue.
> 
> If you want to fix integration tests specifically, you could pin the provider version there: https://github.com/forseti-security/terraform-google-forseti/blob/master/examples/install_simple/main.tf#L18

@morgante Thanks, this was for a quick test. I updated the ticket with more details after testing this morning. I'm going to close this PR and create a new one for the actual fix, but we want to resolve this version issue soon as well. At first I thought the bigquery issue was related to a provider change.

PR #439, opened 2020-01-08
Remove extra dot from tutorial.md

hshin-g:
Remove extra `.` from command in order to prevent CloudShell tutorial from breaking.

### DIFF
--- a/examples/install_simple/tutorial.md
+++ b/examples/install_simple/tutorial.md
@@ -32,7 +32,7 @@ gcloud config set project {{project_id}}
 Run the setup script by providing the Organization ID:
 
 ```bash
-. ../../helpers/setup.sh -p {{project_id}} -o ORG_ID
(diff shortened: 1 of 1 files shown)


gkowalski-google:
Does this resolve an issue? If so, can you link it? The [docs](https://forsetisecurity.org/docs/latest/setup/install.html#create-the-service-account-and-enable-required-apis) and [readme](https://github.com/forseti-security/terraform-google-forseti) will need to be updated if this is removed. The [cleanup script](https://github.com/forseti-security/terraform-google-forseti/blob/master/helpers/cleanup.sh#L42) also relied on an export of this setup script (which is why it was recommended to source it by running with a `.`), but that export has mysteriously disappeared. The cleanup script will need to be updated to remove that dependency.

hshin-g:
No longer able to reproduce! Closing PR.

PR #441, opened 2020-01-09
Fix the Bigquery API issue for release 5.2.0

gkowalski-google:
Update release 5.2.0 for the bigquery change. Update a few remaining places where 5.0.0 was still being used instead of 5.2.0.

### DIFF
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -38,4 +38,4 @@ tags:
 - 'integration'
 substitutions:
   _DOCKER_IMAGE_DEVELOPER_TOOLS: 'cft/developer-tools'
-  _DOCKER_TAG_VERSION_DEVELOPER_TOOLS: '0.1.0'
+  _DOCKER_TAG_VERSION_DEVELOPER_TOOLS: '0.4.6'
--- a/examples/install_simple/main.tf
+++ b/examples/install_simple/main.tf
@@ -37,7 +37,7 @@ provider "random" {
 
 module "forseti-install-simple" {
   source  = "terraform-google-modules/forseti/google"
-  version = "~> 5.2.0"
+  version = "~> 5.1.0"
 
   project_id = var.project_id
   org_id     = var.org_id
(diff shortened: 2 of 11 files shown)


examples/install_simple/main.tf
```diff
@@ -37,7 +37,7 @@ provider "random" {
 
 module "forseti-install-simple" {
   source  = "terraform-google-modules/forseti/google"
-  version = "~> 5.0.0"
+  version = "~> 5.1.0"
```
hshin-g: Is this supposed to be 5.2.0, or is this to fix the tests?

gkowalski-google: 5.2.0 is not released yet, so that will fail. `~> 5.1.0` will use 5.2.0 once it is released. The alternative would be to update this example to point to GitHub instead of the TF repo, but I think the TF repo/versions is better to use.

hshin-g: Ok thanks!

examples/migrate_forseti/tutorial.md
```diff
@@ -160,7 +160,7 @@ to match the region where the CAI GCS bucket is deployed.
 Starting with Forseti Security 2.23, Terraform will manage your server
  configuration file for you.  Configuration options will now be input
  variables that are defined in the Terraform module. Available variables
- and their default values can be found [here](https://github.com/forseti-security/terraform-google-forseti/blob/module-release-5.0.0/variables.tf).
+ and their default values can be found [here](https://github.com/forseti-security/terraform-google-forseti/blob/module-release-5.2.0/variables.tf).
```
hshin-g: Is this the right branch to refer to? I see there is also modulerelease520 that is referenced above.

gkowalski-google: Good catch, thanks. Just committed a change. So many places where the release/branch need to be updated!

build/int.cloudbuild.yaml
```diff
@@ -38,4 +38,4 @@ tags:
 - 'integration'
 substitutions:
   _DOCKER_IMAGE_DEVELOPER_TOOLS: 'cft/developer-tools'
-  _DOCKER_TAG_VERSION_DEVELOPER_TOOLS: '0.1.0'
+  _DOCKER_TAG_VERSION_DEVELOPER_TOOLS: '0.4.6'
```
gkowalski-google: `0.4.6` uses Terraform 0.12.16.

Issue #431, opened 2020-01-02
Using private IPs, helper script service account lacks required IAM permissions


angstwad:
Logging issue quickly as I work; will open PR to remedy or close as required.

```
Error: Error creating GlobalAddress: googleapi: Error 403: Required 'compute.globalAddresses.createInternal' permission for 'projects/durivage-forseti-security/global/addresses/private-ip-ad
dress', forbidden
  on .terraform/modules/forseti/forseti-security-terraform-google-forseti-6a8c52e/modules/cloudsql/main.tf line 42, in resource "google_compute_global_address" "private_ip_address":
  42: resource "google_compute_global_address" "private_ip_address" {
```

blueandgold:
PR is still in code review.  Waiting for user to make recommended changes.

blueandgold:
Will assign to @gkowalski-google as he is doing the code review.

angstwad:
Wilco

PR #442, opened 2020-01-09
Update the module's required IAM permissions

angstwad:
One strategy to deal with the issue presented by #431.  

When creating a Cloud SQL instance with a private IP, looks like the service account's IAM roles lack the permission (compute.globalAddresses.createInternal) required for [this resource's](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/cloudsql/main.tf#L43-L52) creation.


blueandgold:
@angstwad are you able to make the changes that @gkowalski-google is recommending?

angstwad:
@gkowalski-google this is good, whenever you're ready.

PR #444, opened 2020-01-13
Removed simple_example

### LINKED ISSUE #398: Remove the simple_example example
## Story
The simple_example example is mostly redundant of the install_simple, and is no longer needed. The main differences between the two examples is that the simple_example makes Forseti private by default while the install_simple example makes Forseti publish.

There is another issue to provide a [private example](https://github.com/forseti-security/terraform-google-forseti/issues/254) with a [PR](https://github.com/forseti-security/terraform-google-forseti/pull/324).

Consider if the install_simple can be used to provide this private example. If the recommendation is to make Forseti private, then the main example (i.e. install_simple) should be private.

## Acceptance Criteria
- Tests exist for the install_simple example
- A decision has been made about how to show an example of Forseti being private (either use install_simple or a new example)

red2k18:
Resolves #398 
Removed simple example as it is similar to `install_simple`.


test/fixtures/simple_example/main.tf
```diff
@@ -41,11 +41,11 @@ module "bastion" {
   project_id = var.project_id
   subnetwork = var.subnetwork
   zone       = data.google_compute_zones.main.names[0]
-  key_suffix = "_simple_example"
+  key_suffix = "_install_simple"
```
gkowalski-google: This file can be removed as well, there is already a [text fixture](https://github.com/forseti-security/terraform-google-forseti/tree/master/test/fixtures/install_simple) and tests for the install_simple example.

gkowalski-google: Looks like you have done that already, will mark this as resolved and wait for the build to complete.

red2k18: Already removed. I think this happened in the same minute. What a timing!

red2k18: Already removed.

red2k18:
All the tests pass on my local dev env.

gkowalski-google:
The build failed because the `./scripts/wait-for-forseti.sh` file was deleted from the simple_example and the [shared_vpc example fixture](https://github.com/forseti-security/terraform-google-forseti/tree/master/test/fixtures/shared_vpc/scripts) was also relying on this script. I'm fine with just duplicating the script in the shared_vpc example. It should be the same one from the [install-simple fixture](https://github.com/forseti-security/terraform-google-forseti/tree/master/test/fixtures/install_simple/scripts).

PR #446, opened 2020-01-13
Patch update - 5.1.1

hshin-g:
+ Fix Bigquery API name change
+ Update 5.1.1

### DIFF
--- a/examples/install_simple/main.tf
+++ b/examples/install_simple/main.tf
@@ -36,7 +36,8 @@ provider "random" {
 }
 
 module "forseti-install-simple" {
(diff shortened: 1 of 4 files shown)


examples/install_simple/main.tf
```diff
@@ -36,7 +36,8 @@ provider "random" {
 }
 
 module "forseti-install-simple" {
-  source = "../../"
+  source  = "terraform-google-modules/forseti/google"
```
hshin-g: Changed this back from "../../" --> let me know if this should be changed back.

PR #447, opened 2020-01-13
Module Release 5.1.1

hshin-g:
Fix Bigquery API name
Update 5.1.1

### DIFF
--- a/examples/install_simple/main.tf
+++ b/examples/install_simple/main.tf
@@ -14,6 +14,27 @@
  * limitations under the License.
  */
 
+provider "google" {
+  version = "~> 2.11.0"
+  project = var.project_id
+}
+
+provider "local" {
+  version = "~> 1.3"
+}
+
+provider "null" {
+  version = "~> 2.0"
+}
+
+provider "template" {
+  version = "~> 2.0"
+}
+
+provider "random" {
+  version = "~> 2.0"
+}
+
 module "forseti-install-simple" {
   source  = "terraform-google-modules/forseti/google"
   version = "~> 5.0.0"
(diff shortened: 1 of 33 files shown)


examples/install_simple/main.tf
```diff
@@ -14,9 +14,30 @@
  * limitations under the License.
  */
 
+provider "google" {
```
gkowalski-google: I really don't understand why GitHub is showing these provider changes as being added, they already exist in master branch: https://github.com/forseti-security/terraform-google-forseti/blob/master/examples/install_simple/main.tf#L17

examples/install_simple/main.tf
```diff
@@ -14,9 +14,30 @@
...
+}
+
+provider "random" {
+  version = "~> 2.0"
+}
+
 module "forseti-install-simple" {
   source  = "terraform-google-modules/forseti/google"
-  version = "~> 5.0.0"
+  version = "~> 5.1.1"
```
gkowalski-google: This should remain as 5.0.0. 5.1.1 does not exist yet and that's why the lint test failed. The pessimistic operator will use >= 5.0.0 and < 6.0.0, so `~> 5.0.0` will use 5.1.1 once it's released.

examples/migrate_forseti/main.tf
```diff
@@ -40,7 +40,7 @@ provider "random" {
 
 module "forseti" {
   source  = "terraform-google-modules/forseti/google"
-  version = "~> 5.0.0"
+  version = "~> 5.1.1"
```
gkowalski-google: Same as above, leave as 5.0.0

PR #450, opened 2020-01-14
Increased open files limit to fix OSError: [Errno 24] Too many open files 

red2k18:
This PR fixes OSError: [Errno 24] Too many open files error reported in [this issue](https://github.com/forseti-security/forseti-security/issues/3425).

### DIFF
--- a/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
+++ b/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
@@ -155,8 +155,17 @@ systemctl enable --now forseti
 echo "Forseti Startup - Success! The Forseti API server has been enabled and started."
 
 # Increase Open File Limit
-echo "ubuntu soft nofile 32768" | sudo tee -a /etc/security/limits.conf
-echo "ubuntu hard nofile 32768" | sudo tee -a /etc/security/limits.conf
+if grep -q "ubuntu soft nofile" /etc/security/limits.conf ; then
+  echo "Ulimit soft nofile already set."
+else
+  echo "ubuntu soft nofile 32768" | sudo tee -a /etc/security/limits.conf
+fi
+
+if grep -q "ubuntu hard nofile" /etc/security/limits.conf ; then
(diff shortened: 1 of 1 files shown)


modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
```diff
@@ -154,6 +154,10 @@ echo "Forseti Startup - Enabling and starting Forseti service."
 systemctl enable --now forseti
 echo "Forseti Startup - Success! The Forseti API server has been enabled and started."
 
+# Increase Open File Limit
+echo "ubuntu soft nofile 32768" | sudo tee -a /etc/security/limits.conf
```
gkowalski-google: Every time the server resets these lines will get added back into the limits.conf file. It could cause some weird behavior if we try to change this value in the future.

Screen Shot 2020-01-14 at 9 18 49 AM

Instead we can only add these conf changes in if they are not found (duplicate for hard nofile as well):
```
if grep -q "ubuntu soft nofile" /etc/security/limits.conf ; then
  echo "Ulimit soft nofile already set."
else
  echo "ubuntu soft nofile 32768" | sudo tee -a /etc/security/limits.conf
fi
```

red2k18: I have tested this change and we want this values to be added back after the server reset.  Regarding changing this value, we have an [issue](https://app.zenhub.com/workspaces/forseti-5b2c0eaeec7b207d05c80486/issues/forseti-security/terraform-google-forseti/429) for it.

gkowalski-google: We do want the values back after a reset, but without checking if the limit is already added, it will get re-added and be duplicated as seen in my screenshot. I think we should only add the ulimit once. The issue is to add a variable so that this value can be controlled from Terraform.

red2k18: Done

PR #459, opened 2020-01-22
Set the micro version of TF binary in all versions.tf

### LINKED ISSUE #364: Specify the micro version of TF binary in all versions.tf
Right now, the specified version of TF is specified as `0.12`.  There have been issues between micro versions.

We should consider posting the supported TF version (with micro version) in the release notes as well as release communications.

red2k18:
Resolves #364 

Specified the micro version of TF binary in all versions.tf

### DIFF
--- a/examples/external_ip/versions.tf
+++ b/examples/external_ip/versions.tf
@@ -16,5 +16,5 @@
 
 
 terraform {
-  required_version = ">= 0.12"
+  required_version = "0.12.6"
 }
--- a/examples/install_simple/versions.tf
+++ b/examples/install_simple/versions.tf
@@ -15,5 +15,5 @@
  */
 
 terraform {
-  required_version = ">= 0.12"
+  required_version = "0.12.6"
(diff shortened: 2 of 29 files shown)


gkowalski-google:
@blueandgold @hshin-g I know we originally wanted to make this change because of some compatibility issues we saw with Terraform and the Google provider. However, we were able to resolve that by setting provider version dependencies within the Forseti module. This will be a restrictive change and might make it harder for people to use Forseti. Since there haven't been any issues recently with Forseti and Terraform versions, should we focus on upgrading from TF 0.12.6 along with the Terraform Provider versions?

@morgante Always helpful to get your opinion on these changes as well.

morgante:
Yes, I would strongly discourage pinning a specific minor Terraform version. This will make users lives much harder as they won't be able to update to a new Terraform version (even for other modules). No other CFT or reusable modules do this, and HashiCorp strongly discourages it.

PR #457, opened 2020-01-20
Move run_forseti.sh from Forseti repo to Terraform Google Forseti repo

red2k18:
Resolves part of #3313 and #310 

This PR moves [forseti-security/install/gcp/scripts/run_forseti.sh](https://github.com/forseti-security/forseti-security/blob/master/install/gcp/scripts/run_forseti.sh) from Forseti repository to Terraform Google Forseti repository in an effort  to host all deployment scripts at one place (in this repo) and potentially make CV landing easier.

PR #463, opened 2020-01-24
Sync policy library with gsutil rsync

### LINKED ISSUE #461: Mirror policy library directory for GCS sync
## Story
If a user is using Config Validator and the GCS sync to copy policies from GCS to the server VM, if policies are deleted in GCS they are not deleted on the VM.

## Proposed Solution
The [gsutil command](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl#L123) used will copy the polices from GCS to the VM. Before the git-sync feature was added, the policy library was always deleted at server startup, which does not happen anymore. To ensure the local server policy is the same as the GCS policy, we can use `gsutil rsync` instead of `gsutil cp`.

## Acceptance Criteria
- Ensure GCS sync still works by copying down policies
- Delete some policies on GCS and then reset the VM or run the cron job. Ensure that the local policy folder matches the GCS policy.

gkowalski-google:
Switch the gcs sync for the policy library to use gsutil rsync instead of gsutil cp. If a user deletes a policy from GCS, that policy would not get deleted from the local policy library on the Forseti server, so Forseti would continue to scan with it. The gsutil rsync command will ensure that the GCS directory and local policy library are in sync.

I tested by making several changes on the GCS policy, reset the server and ensured that the files deleted were deleted from the local policy library. This PR will ensure the two directories are in sync only during a VM reset. There is another [PR](https://github.com/forseti-security/terraform-google-forseti/pull/457) to move the run_forseti.sh script to the Terraform repo, and will ensure that takes care of syncing the policies as well.

Sample VM logs showing files getting deleted:
```
Jan 24 23:47:57 forseti-server-vm-fe73bca2 startup-script: INFO startup-script: Forseti Startup - Copying Policy Library from GCS.
Jan 24 23:47:57 forseti-server-vm-fe73bca2 startup-script: INFO startup-script: Building synchronization state...
Jan 24 23:47:58 forseti-server-vm-fe73bca2 startup-script: INFO startup-script: Starting synchronization...
Jan 24 23:47:58 forseti-server-vm-fe73bca2 startup-script: INFO startup-script: Removing file:///home/ubuntu/policy-library/policy-library/policies/templates/gcp_always_violates_v1.yaml
Jan 24 23:47:58 forseti-server-vm-fe73bca2 startup-script: INFO startup-script: Removing file:///home/ubuntu/policy-library/policy-library/policies/templates/gcp_allowed_resource_types_v1.yaml
Jan 24 23:47:58 forseti-server-vm-fe73bca2 startup-script: INFO startup-script: Removing file:///home/ubuntu/policy-library/policy-library/policies/templates/gcp_app_service_versions.yaml
```

Resolves #461

### DIFF
--- a/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
+++ b/modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl
@@ -113,14 +113,10 @@ if [ "${policy_library_sync_enabled}" == "true" ]; then
   sudo mkdir -p /etc/git-secret
   sudo gsutil cp -n gs://${storage_bucket_name}/${policy_library_sync_gcs_directory_name}/* /etc/git-secret/
 else
-  # DEPRECATED! Remove once users have migrated over to Git-Sync
   # Download the Newest Config Validator constraints from GCS
   echo "Forseti Startup - Copying Policy Library from GCS."
-
-  # Attempt to download the config-validator policy and gracefully handle the absence
-  # of policy files.  The config-validator is not required for the rest of Forseti
-  # and should not halt installation.
-  gsutil cp -r gs://${storage_bucket_name}/policy-library ${policy_library_home}/ || echo "No policy available, continuing with Forseti installation"
+  sudo mkdir -m 777 -p ${policy_library_home}/policy-library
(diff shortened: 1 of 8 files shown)


gkowalski-google:
@hshin-g Thanks for the review. After you approved I added a test to confirm the policy-library is copied down from GCS correctly, which is passing. PTAL one more time :)

PR #472, opened 2020-01-30
Fix cloudsql password

### LINKED ISSUE #471: Forseti cannot connect to CloudSQL with a password containing multiple @
## Story
By default Forseti will create a random CloudSQL password containing alpha-numeric characters and the special chars `_@`. Having a single `@` seems to not cause any issues, but having 2 `@` causes an error and Forseti cannot connect.

```
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1045, "Access denied for user 'forseti_security_user'@'cloudsqlproxy~' (using password: YES)") (Background on this error at: http://sqlalche.me/e/e3q8)
```

## Proposed Solution
Remove the `@` from possible special characters for the random password.

gkowalski-google:
Remove @ from the list of special characters for the mysql password because it is causing issues. Also, the cloudsql username/password were not being provided to the CloudSQL module so if people provide a username/password then it was not being respected.

Resolves #471

### DIFF
--- a/examples/install_simple/README.md
+++ b/examples/install_simple/README.md
@@ -33,7 +33,6 @@ This configuration is used to simply install Forseti. It includes a full Cloud S
 | forseti-client-storage-bucket | Forseti Client storage bucket |
 | forseti-client-vm-ip | Forseti Client VM private IP address |
 | forseti-client-vm-name | Forseti Client VM name |
-| forseti-cloudsql-user | CloudSQL user |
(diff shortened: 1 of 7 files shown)


hshin-g:
LGTM, just need to resolve conflicts and will approve- thanks!

gkowalski-google:
> LGTM, just need to resolve conflicts and will approve- thanks!

@hshin-g Thanks, resolved!

PR #475, opened 2020-01-30
[FORSETI RELEASE] Update Forseti version to v2.24.1 

red2k18:
[FORSETI RELEASE] Update Forseti version to v2.24.1


red2k18:
Opened #477 instead.

PR #482, opened 2020-02-03
Use internal DNS for client -> server communication

### LINKED ISSUE #295: Client Configuration is not updated when Server IP changes
## Story
Sometimes when the Forseti Server VM is replaced, the Client configuration is not updated to use the new Forseti Server IP. I have seen cases where the Client and Server VM get replaced at the same time, but existing Client configuration still points to the old IP. To fix, users have to manually run `forseti config endpoint {SERVER_IP}:50051` from the client to fix.

## Proposed Solution
Instead of using the server IP address, use the [internal DNS](https://cloud.google.com/compute/docs/internal-dns) of the server. By default GCE instances have a zonal DNS available like `forseti-server-vm-{SUFFIX}.us-central1-c.c.forseti-test123.internal`. Update the client config module to use for the server IP instead of the actual IP.

## Acceptance Criteria
GIVEN: Forseti is deployed on GCE with Terraform
WHEN: Terraform recreates the Forseti Server VM
THEN: The Forseti Client VM always has the correct Server IP

gkowalski-google:
Use the internal DNS name for the Forseti server from the client, and reset the config as part of the Forseti env script. This will ensure the client always has the correct DNS/IP for the server.

Resolves #295

### DIFF
--- a/modules/server/outputs.tf
+++ b/modules/server/outputs.tf
@@ -28,3 +28,8 @@ output "forseti-server-vm-name" {
   description = "Forseti Server VM name"
   value       = google_compute_instance.forseti-server.name
 }
+
+output "forseti-server-vm-internal-dns" {
+  description = "Forseti Server internal DNS"
+  value       = "${google_compute_instance.forseti-server.name}.${google_compute_instance.forseti-server.zone}.c.${var.project_id}.internal"
+}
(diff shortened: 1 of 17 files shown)


modules/server/outputs.tf
```diff
@@ -28,3 +28,8 @@ output "forseti-server-vm-name" {
   description = "Forseti Server VM name"
   value       = google_compute_instance.forseti-server.name
 }
+
+output "forseti-server-vm-internal-dns" {
+  description = "Forseti Server internal DNS"
+  value       = "${google_compute_instance.forseti-server.name}.${google_compute_instance.forseti-server.zone}.c.${var.project_id}.internal"
```
hshin-g: Does not including `${}` work for this variable value?

PR #483, opened 2020-02-04
Expose Cloud SQL instance IP.

xingao267:
This rule makes sure the server VM can still communicating to the SQL
instance when all egress traffic is blocked by default on the server VM.

### DIFF
--- a/modules/server/main.tf
+++ b/modules/server/main.tf
@@ -105,7 +105,7 @@ data "template_file" "forseti_server_env" {
   vars = {
     project_id             = var.project_id
     cloudsql_db_name       = var.cloudsql_module.forseti-cloudsql-db-name
-    cloudsql_db_port       = var.cloudsql_module.forseti-clodusql-db-port
+    cloudsql_db_port       = var.cloudsql_module.forseti-cloudsql-db-port
     cloudsql_region        = var.cloudsql_module.forseti-cloudsql-region
     cloudsql_instance_name = var.cloudsql_module.forseti-cloudsql-instance-name
     cloudsql_db_user       = var.cloudsql_module.forseti-cloudsql-user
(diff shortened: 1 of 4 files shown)


modules/server/main.tf
```diff
@@ -192,6 +192,23 @@ resource "google_compute_firewall" "forseti-server-allow-grpc" {
   depends_on = [null_resource.services-dependency]
 }
 
+resource "google_compute_firewall" "forseti-server-allow-sql-connection" {
```
hshin-g: Might be good to have users choose to opt into this firewall rule by creating a separate variable to enable this rule and add a count of 0 or 1 based on that variable.

xingao267: @gkowalski-google was leaning towards putting this firewall to the customer config where this module is consumed.
@gkowalski-google, do you think gating this firewall rule with a new attribute suggested by @hshin-g is ok? If not, I can discard this PR, and send another one to surface `forseti-cloudsql-instance-ip` all the way to the top level as it will be needed by consumer of forseti module to create the firewall rule dynamically.

xingao267: friend ping~ @gkowalski-google wdyt?

gkowalski-google: Hi @xingao267, sorry for the delay. We were having some internal discussion about this. I think adding the Cloud SQL IP as an output is the preferred way to go. Since Forseti does not need this to operate, and we want the bare minimum configuration to be included in this repo to allow Forseti to work.

xingao267: No worries. Sounds good. I'll modify this PR accordingly. @choonchernlim FYI.

choonchernlim:
@xingao267 @gkowalski-google Thank you. Xin, I'll merge this change into our copy and test it on my end.

PR #498, opened 2020-02-07
Automated roles and APIs needed for Forseti on-GKE deployment

### LINKED ISSUE #500: Automate roles/permissions required to deploy Forseti on-GKE
Automate below roles/permissions required to deploy Forseti on-GKE.

Roles:
1. `roles/container.clusterAdmin`
2. `roles/container.developer`
3. `roles/iam.serviceAccountKeyAdmin`

API:
1. `container.googleapis.com`

red2k18:
Resolves #500 

Automated roles and APIs needed for Forseti on-GKE deployment.

### DIFF
--- a/helpers/setup.sh
+++ b/helpers/setup.sh
@@ -136,6 +136,7 @@ gcloud services enable \
     cloudresourcemanager.googleapis.com \
     serviceusage.googleapis.com \
     compute.googleapis.com \
+    container.googleapis.com \
     --project "${PROJECT_ID}"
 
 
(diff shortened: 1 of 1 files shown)


helpers/setup.sh
```diff
@@ -136,6 +136,7 @@ gcloud services enable \
     cloudresourcemanager.googleapis.com \
     serviceusage.googleapis.com \
     compute.googleapis.com \
+    container.googleapis.com \
```
hshin-g: If this API is not needed for the GCE deployment of Forseti, then might be better to put it under the conditional "if ON_GKE" instead of putting it here.

red2k18: Done

hshin-g:
Thanks for submitting the PR! Before you merge in, you should make sure that these roles/permissions are cleaned up in the cleanup.sh file as well [here](https://github.com/forseti-security/terraform-google-forseti/blob/master/helpers/cleanup.sh).

red2k18:
All the `ON_GKE` roles are granted on the project level and are removed by this [for](https://github.com/forseti-security/terraform-google-forseti/blob/master/helpers/cleanup.sh#L211) block.

PR #502, opened 2020-02-10
Bump google provider version to 3.7.

### LINKED ISSUE #388: Update to the latest version of the Google Provider
## Story
The Terraform Google provider recently [released](https://github.com/terraform-providers/terraform-provider-google/releases) a new major version (3.0.0).  This has caused errors during linting and issues with Forseti deployments.

## Proposal
Update to the new major version of the Google Provider 3.x (currently 3.4) and pin to Forseti versions to a specific version. The goal is to ensure Forseti deployments are stable regardless of new Terraform/Provider versions being released. Forseti is tested against a specific set of dependencies and we want to ensure a good user experience by not running into issues. Remove version specifications from the tests since there is no reason duplicate the versions. It is **strongly recommended** to upgrade to Terraform version 0.12.17 or greater.

Info: With Google Provider version 3.4 breaking changes were introduced between 3.3. This is an example of breaking changes being introduced within a major version, and is a good reason to lock the Forseti Terraform module into specific versions.

## Acceptance Criteria
- The Terraform integration tests pass.

xingao267:
Fixes #388


examples/external_ip/main.tf
```diff
@@ -16,7 +16,7 @@
 
 provider "google" {
   credentials = file(var.credentials_path)
-  version     = "~> 2.11.0"
+  version     = "~> 3.7.0"
```
gkowalski-google: Instead of `~> 3.7.0` can we make this `~> 3.7`? This will need to be made in several other places.

modules/client/main.tf
```diff
@@ -89,7 +89,6 @@ resource "google_compute_instance" "forseti-client" {
   dynamic "network_interface" {
     for_each = local.network_interface
     content {
-      address            = lookup(network_interface.value, "address", null)
```
gkowalski-google: Can you add a comment about why this was removed? What is the impact of removing this? Will this cause any breaking changes? Same with the RTE and server modules.

examples/external_ip/main.tf
xingao267: Done.

modules/client/main.tf
xingao267: Done. There is no impact. `network_ip` is supported to replace `address`, and address was probably kept for backwards compatibility and not removed. `address` was deprecated since google provider 2.0.0, and got removed starting 3.0.0.

And actually, currently it seems neither `address` nor `network_ip` is used at all by the module. So there shouldn't be any impact.

examples/external_ip/main.tf
xingao267: Just to make sure I understand the rationale behind it correctly. Putting 3.7 here instead of 3.7.0 is to leave it open so any version >= 3.7.0 and < 4.0 is supported. According to the documentation regarding ~> here https://www.terraform.io/docs/configuration/terraform.html#specifying-a-required-terraform-version

morgante: Correct. For examples, I usually pin to a minor version (ie. `~> 3.7.0`) while the module itself is less restrictive but either is acceptable.

xingao267: Thanks.

xingao267:
@gkowalski-google @morgante this PR is ready for review again and merge.

chuckwilliams10:
@gkowalski-google --Hi Greg...just wondering when might the Forsetti website get updated to include the new module and the software dependencies content changed...I am referring  to the below content that is online now:    

- Terraform and Plugins

- Terraform 0.12

- Terraform Provider for GCP 2.11.0

- Terraform Provider Templates 2.0

...thanks a ton!!

gkowalski-google:
Hi @chuckwilliams10, sorry for the delay in releasing Forseti v2.25.0 + TF module 5.2.0 + the related website changes. We were waiting at the beginning of this year for a change to allow scanning Org Policy + Access Policy with Config Validator. That change is still in the works by other teams, and we have decided to move forward without being able to fully test scanning those resources with Forseti. Before we publish these new versions, we are updating the GKE examples in the Terraform repo and will do another round of regression testing. Hope that helps!

chuckwilliams10:
@gkowalski-google --thanks for the update..so you guys will release the new Forseti TF module soon after the GKE examples in the TF repo are updated. Thanks a ton for quick responses...any ETA..thanks a ton!  I really, really appreciate it as my team are in the middle of a client engagement that includes deploying the TF Forseti module...thanks a ton!!

chuckwilliams10:
@gkowalski-google --thanks again for the update.  I was just checking any ETA on the new release after the GKE examples in the TF repo are updated...thanks a ton.

Issue #501, opened 2020-02-10
`manage_rules_enabled=false` prevents Forseti service from starting


choonchernlim:
**PROBLEM**

Related to #468...

When setting `manage_rules_enabled` to `false`, the `rules/` dir is not created in the `forseti-server-xxxx` GCS bucket. 

This prevents Forseti service from starting up successfully in server VM:-

```
$ sudo systemctl status forseti
Unit forseti.service could not be found.
$ sudo systemctl restart forseti
Failed to restart forseti.service: Unit forseti.service not found
```

Here's the error from "Serial port 1 (console)" log:- 

```
INFO startup-script: CommandException: No URLs matched: gs://forseti-server-13188a0b/rules
INFO startup-script: Return code 1.
INFO Finished running startup scripts.
```

**PROPOSED SOLUTION**

Ensure `rules/` dir is created in [modules/rules/main.tf](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/rules/main.tf) by adding this code block:-

```
// Creates an empty dir to ensure Forseti service starts if `manage_rules_enabled` is set to false.
resource "google_storage_bucket_object" "empty_rules_dir" {
  count   = var.manage_rules_enabled == false ? 1: 0
  name    = "rules/"
  content = "n/a"
  bucket  = var.server_gcs_module.forseti-server-storage-bucket

  lifecycle {
    ignore_changes = [
      content,
      detect_md5hash,
    ]
  }
}
```

Now, the Forseti service starts successfully in server VM:-

```
$ sudo systemctl status forseti
● forseti.service - Forseti API Server
   Loaded: loaded (/lib/systemd/system/forseti.service; enabled; vendor preset: enabled)
   Active: active (running) since Mon 2020-02-10 16:01:35 UTC; 3min 57s ago
 Main PID: 4084 (forseti_server)
    Tasks: 16 (limit: 4915)
   CGroup: /system.slice/forseti.service
           └─4084 /usr/bin/python3 /usr/local/bin/forseti_server --endpoint [::]:50051 --forseti_db mysql+pymysql:/
Feb 10 16:01:35 forseti-server-vm-13188a0b systemd[1]: Started Forseti API Server.
```

I don't mind creating a PR bug fix for this if you are okay with it. Thank you.

red2k18:
@choonchernlim Thanks for reporting the issue and proposing the solution. Would you be so kind to submit a PR for this?

choonchernlim:
[...]

PR #512, opened 2020-02-19
`manage_rules_enabled=false` should not prevent Forseti service from starting

choonchernlim:
This PR ensures Forseti service will start when `manage_rules_enabled` is set to `false`. 

Per @gkowalski-google, a logging info is added in the startup script, ex:-
```
INFO startup-script: Operation completed over 1 objects/19.8 KiB.
INFO startup-script: Number of rules enabled: 0   <------------------- THIS LINE
INFO startup-script: Forseti Startup - Copying Policy Library from GCS.
```

Test results:-

```
  ✔  server: Forseti Server VM with `manage_rules_enabled=false`
     ✔  Command: `forseti` should exist
     ✔  File /home/ubuntu/forseti-security/rules should be a directory
     ✔  File /home/ubuntu/forseti-security/rules should not have any rules
```

Related: #501


test/fixtures/manage_rules_disabled/main.tf
```diff
@@ -0,0 +1,109 @@
...
+  project_id = var.project_id
+  subnetwork = var.subnetwork
+  zone       = data.google_compute_zones.main.names[0]
+  key_suffix = "_manage_rules_disabled"
+}
+
+#-------------------------#
+# Forseti
+#-------------------------#
+module "forseti-manage-rules-disabled" {
```
gkowalski-google: Thanks @choonchernlim! The build is failing because the Forseti server did not start. The default Forseti version right now is 2.24.0, which is probably hitting the idna issue. The [other examples use the master branch of Forseti](https://github.com/forseti-security/terraform-google-forseti/blob/master/test/fixtures/install_simple/variables.tf#L25), which I recommend here.

GCB errors:
```
null_resource.wait_for_server (remote-exec): Forseti was not ready after 720 seconds!
       
Error: error executing "/tmp/terraform_1299532757.sh": Process exited with status 1
```

choonchernlim: @gkowalski-google My understanding is other branches suffer the same `idna` error because that version isn't pinned. In my current production version, I actually hacked `modules/server/templates/scripts/forseti-server/forseti_server_startup_script.sh.tpl` and forced `pip` to pull `2.8` through `python3 -m pip install 'idna==2.8'` before `setup.py` runs since there's no tagged release from Forseti Security yet.

Regarding that error, it seems I need to increase the wait time, maybe to 900 seconds or higher to allow the provisioning to complete: https://github.com/choonchernlim/terraform-google-forseti/blob/_feature-empty-rules-dir/test/fixtures/manage_rules_disabled/scripts/wait-for-forseti.sh#L21 ... but, I will need to hold off until the next version of Forseti Security is released.

choonchernlim: My bad. I didn't realized my PR got merged into Forseti Security master. Sorry. 

I have been working with too many copies of Forseti. LOL. I increased the timeout, and let's see if that fixes the problem. Thanks.

gkowalski-google: Haha okay thanks, I don't think we need the increased wait timeout but that's ok. Using the master version of Forseti should resolve it. Thanks for all of your recent contributions!

choonchernlim: Thank you for answering all my questions. I totally appreciate it! I'm trying to get the scripts ready to provision Forseti in production in early March. So, I can definitely help in terms of testing and fixing simple problems.

Issue #522, opened 2020-02-25
Add input variable to turn on/off Role scanner


red2k18:
Add input variable to turn on/off [Role scanner](https://github.com/forseti-security/forseti-security/blob/master/samples/scanner/scanners/role_rules.yaml).

PR #526, opened 2020-02-25
Added functionality to enable/disable role scanner

red2k18:
Added functionality to enable/disable role scanner.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -553,6 +553,12 @@ variable "resource_enabled" {
   default     = true
 }
 
+variable "role_enabled" {
+  description = "Role scanner enabled."
+  type        = bool
+  default     = true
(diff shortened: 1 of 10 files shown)


variables.tf
```diff
@@ -553,6 +553,12 @@ variable "resource_enabled" {
   default     = true
 }
 
+variable "role_enabled" {
+  description = "Role scanner enabled."
+  type        = bool
+  default     = true
```
gkowalski-google: Since there aren't any default rules for the Role scanner, should we disable it by default?

red2k18: Yes, disabled Roles scanner.

gkowalski-google:
Resolves #522

PR #531, opened 2020-03-02
Private Deployment of Forseti

kevensen:
The simple_example has been updated to:
1. Deploy with a Cloud NAT
2. Set the Server VM, Client VM, and Cloud SQL database to use private IP
3. Reflect the above changes in the simple_example tutorial


helpers/cleanup.sh
```diff
@@ -142,7 +142,7 @@ gcloud projects remove-iam-policy-binding "${PROJECT_ID}" \
 
 gcloud projects remove-iam-policy-binding "${PROJECT_ID}" \
     --member="serviceAccount:${SERVICE_ACCOUNT_EMAIL}" \
-    --role="roles/compute.networkViewer" \
+    --role="roles/compute.networkAdmin" \
```
gkowalski-google: Should this always be required? If someone is deploying Forseti with public IPs, then this is not required.

helpers/setup.sh
```diff
@@ -178,7 +178,7 @@ gcloud projects add-iam-policy-binding "${PROJECT_ID}" \
 
 gcloud projects add-iam-policy-binding "${PROJECT_ID}" \
     --member="serviceAccount:${SERVICE_ACCOUNT_EMAIL}" \
-    --role="roles/compute.networkViewer" \
+    --role="roles/compute.networkAdmin" \
```
gkowalski-google: Same comment as above in the setup script.

helpers/cleanup.sh
kevensen: It's worth discussing.  If our going in position is that you should always deploy Forseti with private IP's then I say, yes, this is always required.

helpers/setup.sh
hshin-g: Should also put a comment here explaining the purpose of this role, since this is an admin role that gives more permissions.

kevensen:
> Consider making the roles required for a private deployment optional in the setup/cleanup scripts. Update tests to check that the VMs do not have public IPs and that the cloud NAT is setup correctly.

@gkowalski-google Here's the rub.  The `test/setup.sh` creates a Cloud NAT and assigns it to the primary IP ranges on the network where the VM is being deployed.  We can only do this once per region.  In other words, we can't deploy a Cloud NAT in both the `test/setup.sh` script and the `examples/install_simple/main.tf` module.

I have a recommendation.  We take the Cloud NAT out of the example and rather add it to the `<module root>/main.tf` and have everything be private.  I can think of no reason Forseti should be deployed with public IP's at all.  Users will still be able to SSH to the VM's via IAP tunneling with no additional effort.

kevensen:
On second thought, even if we move the router and NAT to the root, we can make the private/public optional.

kevensen:
As an alternative, we leave `test/setup.sh` and `<root module>/main.tf` alone.  We optionally deploy the Cloud NAT in the example (default to yes).  In the case of the test we don't deploy it.  But that means we can't test for it.

test/integration/install_simple/controls/forseti.rb
```diff
@@ -178,6 +199,39 @@
   end
 end
 
+control 'forseti-no-public-ips' do
```
gkowalski-google: (nit) Can you please put this control in a separate file? In an upcoming PR, I was going to move the forseti controls into a shared (symlink) file that can be leveraged by all tests. Since the private IP tests is something that will only be used by the install_simple tests, this would fail.

Issue #310, opened 2019-10-14
Migrate remaining scripts from forseti-security repo


kevensen:
The following scripts should be maintained in this repo and deployed to the server GCE VM.

- [x] [forseti-security/install/gcp/scripts/initialize_forseti_services.sh](https://github.com/forseti-security/forseti-security/blob/master/install/gcp/scripts/initialize_forseti_services.sh)
- [x] [forseti-security/install/gcp/scripts/run_forseti.sh](https://github.com/forseti-security/forseti-security/blob/master/install/gcp/scripts/run_forseti.sh)

See https://github.com/forseti-security/forseti-security/issues/3313

PR #536, opened 2020-03-05
Migrate initialize_forseti_services.sh script from Forseti repo to Terraform Google Forseti repo

red2k18:
Resolves part of #310


modules/server/templates/scripts/forseti-server/initialize_forseti_services.sh.tpl
```diff
@@ -0,0 +1,164 @@
...
+echo "Cloud SQL Instance Connection string: ${cloudsql_connection_name}"
+echo "SQL port: ${cloudsql_db_port}"
+echo "Forseti DB name: ${cloudsql_db_name}"
+
+if ! [[ -f ${forseti_server_conf_path} ]]; then
+    echo "Could not find the configuration file: ${forseti_server_conf_path}." >&2
+    exit 1
+fi
+
+# We had issue creating DB user through deployment template, if the issue is
```
gkowalski-google: (nit) This comment is no longer relevant.

red2k18:
@gkowalski-google Thanks for taking a look! Please let me know when you complete the review. Would like to address any other comment you have at once so integration tests are not executed multiple times (would like to avoid running into the org IAM issues that I ran into last week). Also, it would be great if you could prioritize the review.

PR #540, opened 2020-03-08
Ability to install Stackdriver monitoring in server VM

choonchernlim:
This PR introduces 1 new variable to [install Stackdriver monitoring](https://cloud.google.com/monitoring/agent/install-agent#agent-install-debian-ubuntubash) on VM startup and ensures the service account has the correct permission to use it.

```
module "forseti" {
  source = "..."

  // 1 new variable (optional, default = `false`)
  server_monitoring_enabled = true
}
```

Test results:-

```
  ✔  server_gcp: Forseti Server VM with `server_monitoring_enabled=true`
     ✔  Project IAM Binding roles/monitoring.metricWriter should exist
     ✔  Project IAM Binding roles/monitoring.metricWriter members should include "serviceAccount:forseti-server-gcp-89f3e9c6@ci-forseti-serv-qpyo.iam.gserviceaccount.com"
  ✔  server_ssh: Forseti Server VM with `server_monitoring_enabled=true`
     ✔  Service stackdriver-agent should be installed
     ✔  Service stackdriver-agent should be enabled
     ✔  Service stackdriver-agent should be running
```


choonchernlim:
@gkowalski-google I could use another pair of eyes from CFT side to verify the correctness in installing Stackdriver monitoring agent in this PR.

This PR uses the new approach of installing the agent:-
```
curl -sSO https://dl.google.com/cloudagents/add-monitoring-agent-repo.sh
sudo bash add-monitoring-agent-repo.sh
sudo apt-get update
sudo apt-get install -y stackdriver-agent
sudo service stackdriver-agent start
```
... but, from my end, we are starting to see lots of `infinite value` errors in the log after the agent is installed.

Due to time crunch, I reverted my implementation back to the deprecated way  (not in this PR):-
```
curl -sSO https://dl.google.com/cloudagents/install-monitoring-agent.sh
bash install-monitoring-agent.sh
```

But I could use some help in figuring out what causes the `infinite value` errors from that agent.

Thank you.

hshin-g:
@choonchernlim Would you be able to resolve the branch conflicts here?

red2k18:
@choonchernlim Would you be able to resolve the branch conflicts here?

red2k18:
@gkowalski-google Would you be able to review this PR?

red2k18:
@choonchernlim , please resolve the branch conflicts so we can merge in this PR as well.

red2k18:
@choonchernlim , there are some lint errors that need to be fixed. You can run `make docker_test_lint` to see the lint errors on your local.

gkowalski-google:
Hi @choonchernlim, thanks for this PR. The tests are failing because of `Error: Error waiting for instance to create: Quota 'CPUS' exceeded.  Limit: 72.0 in region us-central1.`. We have too many GCE instances created for the test project. Tests are already taking over 2 hours to run, and I think we need to decide about how to handle testing for some of these parameters going forward. While it is nice to have these tests, they are expensive and I don't think we can continue adding tests for every parameter.

I also don't have a great solution. The best I can think of right now is to start combining some of these tests into a single fixture. I'd like to leave the testing for the common examples as is.

What if we create a new test project and start combining some of these tests. For example, there are 2 VMs created to test that `managed_rules_disabled` is working correctly. Instead we can do a single deployment (for now) of Forseti into this new project and use that to test managed rules disabled and this new Stackdriver monitoring feature. As long as the features are compatible, I think it's OK. Perhaps the shielded vm testing could be done on this deployment as well. What do you think?

choonchernlim:
I agreed the tests could probably be consolidated into just a few fixtures. Right now, it takes too long to run all the fixtures. Unfortunately, I don't have much cycles at the moment to do this refactoring... too many datelines lately.

red2k18:
@gkowalski-google @choonchernlim Should we close this PR if work will not be done in the near future?

gkowalski-google:
@red2k18 Let's keep it open for now. I will see if I can make the changes before the next release.

PR #546, opened 2020-03-12
Support Bring-Your-Own Service Accounts.

### LINKED ISSUE #243: Installer should support specifying existing service accounts
Install should support specifying existing Service Accounts to be used for forseti-client and forseti-server.

Since IAM is normally delegated to a different team than the one deploying Forseti, this would reduce the number of Service Account request/approval/provisioning cycles when the individual deploying Forseti already has an account with the necessary permissions but lacks SetIAM permissions to setup the installer generated Service Accounts.

xingao267:
Fixes #243

### DIFF
--- a/test/fixtures/custom_service_accounts/main.tf
+++ b/test/fixtures/custom_service_accounts/main.tf
@@ -1,0 +1,39 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+resource "google_service_account" "forseti_server" {
+  account_id   = "my-custom-server-sa"
+  project      = var.project_id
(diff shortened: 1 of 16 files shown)


test/fixtures/custom_service_accounts/main.tf
```diff
@@ -0,0 +1,39 @@
...
+
+module "custom-service-accounts" {
+  source = "../../.."
+
+  project_id             = var.project_id
+  org_id                 = var.org_id
+  domain                 = var.domain
+  network                = var.network
+  subnetwork             = var.subnetwork
+  server_service_account = google_service_account.forseti_server.email
```
xingao267: I got error because 
```
Error: Invalid count argument
       
         on ../../../modules/client_iam/main.tf line 33, in resource "google_service_account" "forseti_client":
         33:   count        = (var.client_enabled && var.client_service_account == "") ? 1 : 0
       
       The "count" value depends on resource attributes that cannot be determined
       until apply, so Terraform cannot predict how many instances will be created.
       To work around this, use the -target argument to first apply only the
       resources that the count depends on.
       
       
       Error: Invalid count argument
       
         on ../../../modules/server_iam/main.tf line 56, in resource "google_service_account" "forseti_server":
         56:   count        = var.server_service_account == "" ? 1 : 0
       
       The "count" value depends on resource attributes that cannot be determined
       until apply, so Terraform cannot predict how many instances will be created.
       To work around this, use the -target argument to first apply only the
       resources that the count depends on.
```
It seems `count` does not like `google_service_account.forseti_server.email` as the value cannot be determined before the resource is applied? Is there suggested ways to get around this? I tried another approach to directly hardcode the service account emails here but I can't enforce the dependency using `depends_on` here because it's a module.

xingao267:
@gkowalski-google ready for review.

PR #581, opened 2020-04-30
Updates for TF 0.12.20 and Readme cleanup

### LINKED ISSUE #557: Upgrade automated tests to use Terraform 0.12.20
## Story
With the v6.0.0 module release, the Forseti module will be tested against TF 0.12.20 and the team will not support using the module with versions < 0.12.20.

## Solution
Update the tests to use TF 0.12.20, which is tag version `0.7.5` of the dev tools image. Update any documentation.

gkowalski-google:
Update module to support TF 0.12.20. Update readme with some general cleanup and info specific to TF 0.12.20.

Resolves #557

### DIFF
--- a/test/setup/main.tf
+++ b/test/setup/main.tf
@@ -64,7 +64,6 @@ module "forseti-host-project" {
     "sqladmin.googleapis.com",
     "logging.googleapis.com",
     "monitoring.googleapis.com",
-    "servicenetworking.googleapis.com",
   ]
 }
 
--- a/test/setup/main.tf
+++ b/test/setup/main.tf
@@ -64,6 +64,7 @@ module "forseti-host-project" {
     "sqladmin.googleapis.com",
     "logging.googleapis.com",
     "monitoring.googleapis.com",
+    "servicenetworking.googleapis.com",
   ]
 }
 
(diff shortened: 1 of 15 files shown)


test/setup/main.tf
```diff
@@ -64,6 +64,7 @@ module "forseti-host-project" {
     "sqladmin.googleapis.com",
     "logging.googleapis.com",
     "monitoring.googleapis.com",
+    "servicenetworking.googleapis.com",
```
hshin-g, line 67: Does `main.tf` also need this API listed?

gkowalski-google: I have updated the cloudsql module to handle this logic.

gkowalski-google: Nevermind, I only fixed the shielded vm and shared vpc tests. The install simple test is done a little differently than would typically be deployed, so to handle that I am putting this service back in the test setup. I will leave [this](https://github.com/forseti-security/terraform-google-forseti/pull/581/files#diff-e68da704e977beb85b34e6adfa2bcebbR43) in which will handle the common scenarios where Forseti is deployed into a project that is different from where the network is setup. However, since the install simple tests are deployed into the same project as the network, it doesn't work there. To really fix it, we should change the install simple example test to be done where the network is setup in a different project.. and I don't have the heart to do that in this PR.

gkowalski-google: @hshin-g Updated PTAL.

PR #616, opened 2021-01-08
Allow Cloud SQL deletion

e-koma:
## What Pull Reqeust
parameterized Cloud SQL's `deletion_protection`.

## Why
Cloud SQL( `google_sql_database_instance` ) is default  `deletion_protection = "true"`,  which makes it impossible to destroy terraform. I actually encountered the following situations.

1. terraform apply fails when Cloud SQL's destructive parameters are set.
2. In the situation like forseti setting did not works well(Cloud SQL starts, but GKE does not start well), I want to delete Cloud SQL, but  cannot delete it.
3. even if we want to run forseti as a trial, but after trying, we cannot delete Cloud SQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ For this module to work, you need the following APIs enabled on the Forseti proj
 | cloudsql\_db\_password | CloudSQL database password | string | `""` | no |
 | cloudsql\_db\_port | CloudSQL database port | string | `"3306"` | no |
 | cloudsql\_db\_user | CloudSQL database user | string | `"forseti_security_user"` | no |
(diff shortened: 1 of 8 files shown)


e-koma:
@googlebot I signed it!

e-koma:
I now understand 
this issue cause  terraform-provider-google version is above `3.44` in my local environment.

`deletion_protection` option is set by default true after google provider version `3.44`,  and cannot delete it.
https://github.com/hashicorp/terraform-provider-google/blob/master/CHANGELOG.md#3440-october-19-2020


but forseti examples google provider are below 3.44.
So this issue is not relevant in the current forseti's google provider version. 
I'm going to rewrite my local configuration to run it.

Thank you for your patience.

PR #617, opened 2021-01-13
Replace using of WhiteList with AllowList and BlackList with DenyList.

e-koma:
I completely agree with the deletion of all blacklist / whitelist.
https://www.cnet.com/news/twitter-engineers-replace-racially-loaded-tech-terms-like-master-slave/

I replaced using of WhiteList with AllowList and BlackList with DenyList.


e-koma:
What test was failed in CloudBuild ? 
In my local environment, I don't get any test errors.

I only get the error of deleting Cloud SQL after some tests are completed. 
```
failed to delete instance because deletion_protection is set to true.
```
but, This error occurs when the terraform google provider version is 3.44 or higher(deletion_protection is default true),  
so I consider it normal.

PR #626, opened 2021-05-10
Fix shielded vm test

gkowalski-google:
Remove the boot images for the client and server from the shielded vm test. These images no longer exist.


gkowalski-google:
More changes to get head back into a working and testable state.

EricBeach (approved):

PR #627, opened 2021-05-10
Update minimum version of the Google provider to 3.52

gkowalski-google:
This is needed to support the deletion_protection argument for CloudSQL and using the last version of the Cloud NAT module which works with Terraform 0.12. The lint tests will pass, and int test will still fail on not being able to delete the test VMs. The next PR will actually get the tests all passing 👍

### DIFF
--- a/examples/external_ip/main.tf
+++ b/examples/external_ip/main.tf
@@ -16,7 +16,7 @@
 
 provider "google" {
   credentials = file(var.credentials_path)
-  version     = "~> 3.7"
+  version     = "~> 3.52"
 }
 
 provider "null" {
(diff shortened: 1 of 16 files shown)


EricBeach (approved):

gkowalski-google:
Hi @morgante, just thought I'd check with you to see if you have any concern raising up the minimum required version of the Google provider here. This is being done so that we can add the `deletion_prevention` variable to allow for the Cloud SQL instance to be deleted.

morgante:
> Hi @morgante, just thought I'd check with you to see if you have any concern raising up the minimum required version of the Google provider here. This is being done so that we can add the `deletion_prevention` variable to allow for the Cloud SQL instance to be deleted.

No problem for me, go ahead.

PR #628, opened 2021-05-11
Add deletion_protection for Cloud SQL instance

gkowalski-google:
Add support for deletion_protection, so that the CloudSQL database instance can be deleted by tests and for users who are trying to destroy the Forseti deployment.


e-koma:
could you please re open it ?

PR #632, opened 2021-06-23
Fix/gcs terraform

CpBulletSponge:
Updated terraform code to remove gcs 'bucket_policy_only' which is to be deprecated to use 'uniform_bucket_level_access'


CpBulletSponge:
@googlebot I signed it!

gkowalski-google:
I'm not opposed to this change, but it is a breaking change for existing users who are using the main branch to deploy Forseti.

gkowalski-google:
An alternative is to keep the `bucket_policy_only` var along with the new `uniform_bucket_level_access`. Use a local var to decide which one to use.

